PR #40, opened 2014-08-18
Reverse frames since stacktrace.js gives them to us in most recent call first order

coryvirok:


test/notifier.test.js
```diff
@@ -1057,6 +1057,38 @@ describe("Notifier._buildPayload()", function() {
...
+    } catch (e) {
+      err = e
+    }
+    var payload = notifier._buildPayload(new Date(), 'error', 'frame order test', TK(err));
+    expect(payload.data.body.trace.frames).to.be.an('array');
+
+    var numFrames = payload.data.body.trace.frames.length;
+    expect(payload.data.body.trace.frames[numFrames - 3].method).to.equal('first');
+    expect(payload.data.body.trace.frames[numFrames - 2].method).to.equal('second');
+    expect(payload.data.body.trace.frames[numFrames - 1].method).to.equal('last');
```
coryvirok: Hrm, is this even correct? I'd expect the last frame to have "first".

jondeandres: @coryvirok this PR is yours no? :smile:

jondeandres: @coryvirok I've been looking at this and I've opened a PR to be merge in the tracekit replacement branch: https://github.com/rollbar/rollbar.js/pull/82.

I would expect `payload.data.body.trace.frames[0].method` to be `last`. Are you sure that actually is not working correctly?

jondeandres: Oh @coryvirok!! I think it's ok.

Our test say `should have the most recent frame last`. So in the context of the error, the **most** recent frame is `last`. Since the frames are:

```javascript
last()
second()
first()
// here all the other frames
```

So I think it's correct.

If we want to release this with webpack and tracekit replacement we can close this an merge https://github.com/rollbar/rollbar.js/pull/82 into `replace-tracekit`.

jondeandres: https://github.com/rollbar/rollbar.js/pull/82

coryvirok: Thanks. My brain must have been fried yesterday. Let's merge #82 and close this.

PR #68, opened 2015-02-06
Return notifier instance from `init`

sandstrom:
I'm trying to understand if I've totally misunderstood something (please let me know if I have!).

But I think the instance must be returned, otherwise there isn't a way to send errors to rollbar via `log` (when the global errorhandler isn't used, which is often the case for larger frameworks).


sandstrom:
Solved by 1.2.0

PR #77, opened 2015-04-26
Remove the 1s repeating timer that processed the error queue

cowsrule:
Using a repeating timer to call Notifier.processPayloads every second is unnecessary. The error reporting queue can be notified when there is data present and schedule time to send data to the remote server instead of polling to see if there is data. The processing of error payloads is left async using setTimeout to avoid doing a large amount of work right when reporting an error and allows for batching.

### DIFF
--- a/src/notifier.js
+++ b/src/notifier.js
@@ -631,6 +631,8 @@ Notifier.prototype._enqueuePayload = function(payload, isUncaught, callerArgs, c
 
   if (!!this.options.enabled) {
     window._rollbarPayloadQueue.push(payloadToSend);
+
+    _notifyPayloadAvailable();
   }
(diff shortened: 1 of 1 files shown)


cowsrule:
This change may need additional work to support cases where the user goes offline with errors stuck in the queue which would not get removed unless the queue. Granted I don't actually see any code that would retry sending an error on an XHR failure so this may already be an issue...

coryvirok:
Thanks for the PR, @cowsrule. Our team will take a look.

jondeandres:
@cowsrule we've closed this PR using this one https://github.com/rollbar/rollbar.js/pull/85.

Thanks for this!

PR #78, opened 2015-04-27
Add distribution that omits the custom JSON implementation

brianr:
This fixes Content-Security-Policy violations in environments that disallow `unsafe-eval` (for example, Chrome extensions).


coryvirok:
nice :+1:

PR #80, opened 2015-05-07
Webpack

jondeandres:
I've configured webpack to create 3 different bundles `rollbar.js`, `rollbar.nojson.js` and `rollbar.umd.js`, check the entry points in `src/bundles`

They are created following the concat configuration we have in `Gruntfile.js` but using `require` where needed. It's been a bit tricky configure the JSON library to use. 

In `dist/bundles/examples` there are some HTML examples of use. I've checked `rollbar.js` and `rollbar.nojson.js` with a script tag and `rollbar.umd.js` with requirejs.


webpack.config.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  entry: {
+    rollbar: './src/bundles/rollbar.js',
+    'rollbar.nojson': './src/bundles/rollbar.nojson.js',
+    'rollbar.umd': './src/bundles/rollbar.umd.js'
```
coryvirok: I think customers will want a version of the UMD without JSON and one with. 

I'd expect the build process to add the custom JSON implementation for one build but not the other. The custom JSON code would override `Rollbar._json` or something similar.

e.g.
```js
module.exports = [
    // Rollbar with custom JSON
    {
        name: 'with-json',
        entry: {
            'rollbar': './src/bundles/rollbar.js',
            'json': './src/bundles/json2.js'
        },
        output: {
            library: 'Rollbar',
            path: './dist/bundles/',
            libraryTarger: 'umd.json'
        },
        failOnError: true
    },

    // Rollbar without custom JSON
    {
        name: 'without-json',
        entry: {
            'rollbar': './src/bundles/rollbar.js'
        },
        output: {
            library: 'Rollbar',
            path: './dist/bundles/',
            libraryTarger: 'umd.no-json'
        },
        failOnError: true
    }
];
```

webpack.config.js
```diff
@@ -0,0 +1,15 @@
...
+  },
+  output: {
+    library: 'Rollbar',
+    path: './dist/bundles/',
```
coryvirok: Does this mean that we will no longer be placing the distributed, (minified, etc.) files into `dist/`? I think we should keep them there.

Gruntfile.js
```diff
@@ -13,6 +15,9 @@ module.exports = function(grunt) {
 
   grunt.initConfig({
     pkg: pkg,
+    webpack: {
+      options: webpackConfig
+    },
     jshint: {
       options: {
         globals: {
```
coryvirok: There's lots of stuff below that we won't need anymore. E.g. the 'concat/amd/commonjs' targets.

coryvirok:
It's coming along. Nice work so far.

webpack.config.js
jondeandres: @coryvirok yah, I'll change this. This was only to don't have conflicts with the existing ones while the development process.

jondeandres:
@coryvirok I've done some important changes in the PR.

Now, tracekit and json2 are minified before execute webpack, since they are injected as they are, but as an string, so it's not minified later by uglify.

I've added the bundle for the snippet. `shimload.js' content goes to the bundle, and `shim.js` is required from there. 

Now we have just 3 bundle files: normal or browser tag, umd and the snippet. For normal and umd I've configured webpack with to expose `__USE_JSON__` with two different values: `true` and `false`. So we can setup the JSON or not depending on that value. The use of this feature in webpack is quite transparent and simple.

Let me know how looks for you.

webpack.config.js
```diff
@@ -0,0 +1,68 @@
...
+      rollbar: './src/bundles/rollbar.js',
+      'rollbar.umd': './src/bundles/rollbar.umd.js'
+    },
+    output: {
+      library: 'Rollbar',
+      path: outputPath,
+      filename: '[name].js',
+      libraryTarget: 'umd'
+    },
+    plugins: [new webpack.DefinePlugin(jsonDefines)],
```
coryvirok: Oh, cool!

webpack.config.js
```diff
@@ -0,0 +1,68 @@
+var webpack = require('webpack');
+
+var outputPath = './dist/';
+
+var jsonDefines = {
+  __USE_JSON__: true
+};
+
+var NoJsonDefines = {
```
coryvirok: nit: lowercase -> `noJsonDefines`

src/bundles/rollbar.js
```diff
@@ -0,0 +1,34 @@
+var notifier = require('../notifier');
```
coryvirok: Is there any reason to keep this file?

Gruntfile.js
```diff
@@ -210,11 +188,12 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('grunt-bumpup');
   grunt.loadNpmTasks('grunt-tagrelease');
   grunt.loadNpmTasks('grunt-saucelabs');
+  grunt.loadNpmTasks('grunt-webpack');
 
-  grunt.registerTask('build', ['replace', 'jshint', 'concat', 'uglify']);
+  grunt.registerTask('build', ['replace', 'jshint', 'uglify:prewebpack', 'webpack', 'uglify:dist']);
```
coryvirok: I believe you can also have Webpack do the uglification for you: http://webpack.github.io/docs/list-of-plugins.html#uglifyjsplugin

dist/rollbar.umd.nojson.js
```diff
@@ -0,0 +1,1460 @@
...
+/******/ ([
+/* 0 */
+/***/ function(module, exports, __webpack_require__) {
+
+	var globalnotifier = __webpack_require__(1);
+	
+	function setupJSON() {
+	  var JSONObject = JSON;
+	
+	  if (false) {
```
coryvirok: It would be nice if Webpack didn't include this since it will never evaluate it. Do you know if there is a way to do this? Is the plugin we are using, (https://github.com/rollbar/rollbar.js/pull/80/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR5)  the right tool for this job?

src/bundles/rollbar.snippet.js
```diff
@@ -0,0 +1,10 @@
+var RollbarShim = require('../shim').Rollbar;
+var snippetCallback = require('../snippet_callback');
+
+var defaultRollbarJsUrl = '//d37gvrvc0wt4s1.cloudfront.net/js/v1.2/rollbar.min.js';
```
coryvirok: This URL should come from the package.json list of defaults.

dist/rollbar.umd.nojson.js
brianr: I believe Webpack will automatically remove this via dead-code elimination when it generates the minified build.

jondeandres: It seems that uglify removes this block of code. @coryvirok is ok?

Gruntfile.js
jondeandres: @coryvirok we can replace `uglify:dist` task with the webpack uglify plugin, but seems that the plugin only creates the minified version. Is ok for you?

`uglify:prewebpack` task cannot be removed, we need to minify json2 and tracekit before use them in webpack.

PR #83, opened 2015-05-26
Replace http protocol to https on rollbarJsUrl config option.

jondeandres:


coryvirok:
Seems like a bit of overkill for doing this. Can you think of anything else we would want to normalize in the config?

jondeandres:
@coryvirok not actually, I can change it, but I thought it was a good ida having a isolated module to take care about the config format and others.

I'll change this.

jondeandres:
@coryvirok I've replaced all that code with this https://github.com/rollbar/rollbar.js/commit/534f1e142fbcf6c8fe28b5b295fa8c04bf3428ac.

coryvirok:
LGTM

jondeandres:
@coryvirok what do you think?

@brianr for me that's fine, but I thought the problem exists when users sets by hand `rollbarJsUrl`.

jondeandres:
The bad point I find is that this can be a problem developing mox, since the development server doesn't have ssl enabled no?

brianr:
@jondeandres The purpose of this is to work around the issues we see where some proxies block OPTIONS requests (needed for CORS) when sent over HTTP. So forcing everything to go over HTTPS makes rollbar.js work out of the box for more clients.

For developing mox itself, you can still set `rollbarJsUrl` yourself to whatever you want. Though not if this PR is merged as-is :) That's my point - we should just change the default url to be https instead of schemaless.

jondeandres:
Ok! I'll change this.

jondeandres:
@brianr changed in https://github.com/rollbar/rollbar.js/commit/c2df60ffad9e128c4b0c20c8e96706b29f6a6228.

PR #85, opened 2015-05-28
Remove repeated timer

jondeandres:


src/notifier.js
```diff
@@ -868,19 +870,29 @@ function _guessErrorClass(errMsg) {
 
 var payloadProcessorTimeout;
 Notifier.processPayloads = function(immediate) {
-  if (!payloadProcessorTimeout || immediate) {
-    _payloadProcessorTimer(immediate);
+  if (immediate) {
+    _deferredPayloadProcess();
+  }
+  else if (!payloadProcessorTimeout) {
```
brianr: Looks like this duplicated the test in `_notifyPayloadAvailable` and could be replaced with:

```
} else {
  _notifyPayloadAvailable();
}
```

PR #87, opened 2015-05-28
Change argument name in Stack() to fix uglify bug.

jondeandres:
The minified version of the buggy function is this:

```javascript
function o(n) {
            function e() {
                var e = [];
                try {
                    e = l.parse(n)
                } catch (t) {
                    e = []
                }
                for (var r = [], s = 0; s < e.length; s++) r.push(new a(e[s]));
                return r
            }
            return {
                stack: e(),
                message: n.message,
                name: n.name
            }
        }
```

### DIFF
--- a/src/error_parser.js
+++ b/src/error_parser.js
@@ -27,12 +27,12 @@ function Frame(stackFrame) {
   return data;
 };
 
-function Stack(e) {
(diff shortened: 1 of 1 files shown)


coryvirok:
:+1: LGTM

PR #91, opened 2015-06-10
Added polyfills for IE8+ and now using a forked version of error-stac…

coryvirok:
…k-parser @jondeandres cc @brianr


jondeandres:
very nice @coryvirok.

PR #93, opened 2015-06-10
Prepare 1.3 version

jondeandres:
I've changed `README.md` with the new snippet and UMD information. Also I've gone through all the examples and I've fixed and completed them.

I've also added `examples/snippet.html` and `examples/script.html` to show functional examples using the snippet and script tag, requirejs is covered in `examples/requirejs/test.html`.

@coryvirok we have a problem with this example, https://github.com/rollbar/rollbar.js/tree/master/examples/standalone. I was thinking that all of that could be replaced just suggesting paste the whole UMD file in a `script` tag. I don't know why the shim is needed in that example.


coryvirok:
For the standalone example, we no longer need the shims because we can just include the rollbar.umd.js script in a script tag and then use `init()`:
```html
<html>
  <head>
    <script type="text/javascript" src="rollbar.umd.js"></script>
    <script>
      var _rollbarConfig = {...};
      Rollbar.init(_rollbarConfig);
    </script>
  </head>
  ...
</html>
```

jondeandres:
@coryvirok I'll remove the standalone example then.

PR #97, opened 2015-06-19
Fix UMD section in README.md.

jondeandres:


piotrd:
Uhm, not entirely: `You can find examples of use in [examples].` the link to examples still gives 404.

jondeandres:
Oh! Yes @piotrd, I was confused with Github releases page and our `/release` path. Now is fixed, thanks!

PR #99, opened 2015-06-23
Don't call shim's onerror and use the original one.

jondeandres:
This prevents the shim to send reports when full notifier is already loaded.

### DIFF
--- a/src/globalnotifier.js
+++ b/src/globalnotifier.js
@@ -45,11 +45,19 @@ wrapper.init = function(config, parent) {
 
   if (config.captureUncaught) {
     // Set the global onerror handler
-    var old = window.onerror;
+    var oldOnError;
+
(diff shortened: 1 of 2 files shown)


src/globalnotifier.js
```diff
@@ -45,11 +45,19 @@ wrapper.init = function(config, parent) {
 
   if (config.captureUncaught) {
     // Set the global onerror handler
-    var old = window.onerror;
+    var oldOnError;
+
+    // If the parent, probably a shim, stores a oldOnError, use that so we don't
+    // send reports twice.
+    if (parent && typeof parent.oldOnError !== 'undefined') {
```
brianr: Would it make sense to put 'rollbar' in that variable name somewhere, to make sure it really was us? e.g. `parent._rollbar_oldOnError` or something.

jondeandres: Ok, no proboem.

PR #102, opened 2015-06-26
Remove umd as the default target in the snippet and update tests. @jo…

coryvirok:
…ndeandres


jondeandres:
@coryvirok this is nice, thanks for the job. maybe you can change the example for snippet in `examples/snippet.html`.

PR #112, opened 2015-07-23
Fix issue with selenium throwing permission denied to access wrap

robertosanchez:
This adds the same fix as in `shim.js` to prevent the wrap function to crash when running with selenium


robertosanchez:
closing and re-opening to rerty travis build

PR #116, opened 2015-07-27
Use always HTTPS for request to our API. @brianr @coryvirok

jondeandres:


dist/rollbar.js
```diff
@@ -254,10 +254,7 @@
 	  // when the rate limit is reached.
 	  _topLevelNotifier = _topLevelNotifier || this;
 	
-	  var protocol = window.location.protocol;
-	  if (protocol.indexOf('http') !== 0) {
-	    protocol = 'https:';
-	  }
+	  var protocol = 'https';
```
brianr: Looks like this variable is only used one place, the very next line. Any reason not to inline it in the next line?

brianr:
A couple other issues here:

- The change should be made in the `src/` directory. `dist/rollbar.js` is a generated file.
- Build is failing (looks unrelated to this change, but we should get it passing before merge)

dist/rollbar.js
jondeandres: I'll change it, ok.

jondeandres:
@brianr oh yes, seems I pushed the change in dist/ instead of src/. Let me change it.

jondeandres:
@brianr @coryvirok https://github.com/rollbar/rollbar.js/commit/8517ac44c5ab71d150789bbf76d89115bfb2761d

PR #117, opened 2015-07-27
Fix selenium issues

robertosanchez:
This adds the same fix as in `shim.js` to prevent the wrap function to crash when running with selenium.

Also, I updated `package.json` based on #116 to allow the build in travis to run correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "mocha": "^1.21.4",
     "mocha-phantomjs": "^3.5.0",
     "node-libs-browser": "^0.5.2",
-    "phantomjs": "^1.9.10",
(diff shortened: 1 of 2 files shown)


robertosanchez:
@brianr @coryvirok any thought on this change? Does it look like a correct approach?

brianr:
@robertosanchez Thanks for this! I'll defer to @coryvirok for the review on this change.

coryvirok:
LGTM. I'll merge soon. I have another change I'd like to also get into the next release so I'll release a new version after that.

robertosanchez:
Great! thanks!

Issue #63, opened 2014-12-19
If err.stack can't be parsed, the reported stack trace is not helpful


brianr:
If you do the following:

``` js
var err = new Error("hello");
err.stack = "not a valid stack";
Rollbar.error(err);
```

Then it will appear in Rollbar like this: https://rollbar.com/Rollbar/ember-example/items/3/ (public link)

This appears to be happening because if TraceKit (which we use to parse the stack trace) can't parse the stack, it uses the current stack trace instead... but this isn't very helpful. Instead, rollbar.js should:
- include the raw value of `err.stack` (i.e. the original stack string from the browser, or in the example above where it's clobbered, the string `"not a valid stack"`)
- include a note about how it couldn't be parsed

coryvirok:
@brianr Is this still an issue?

brianr:
@coryvirok Looks like yes -- the behavior is slightly different in 1.3.0, but we're still losing the original stack when we can't parse it.

After executing the example code from the description, the result is this: https://rollbar.com/Rollbar/rollbar.js/items/350/occurrences/6000486028/

coryvirok:
Thanks. I'll file a story to send along the original stack trace with every error so we can both debug this and show it in the UI.

formigarafa:
I believe this could be also related to subclassing Errors.

`Util.typeName` at https://github.com/rollbar/rollbar.js/blob/master/src/util.js#L31 fails to recognize `Error` subclasses as errors.

And this helper function is used to determine which argument being passed is the error object from which the stack trace is inferred.

As a proof of concept I did the following:
```
var originalObjectToString = window.Object.prototype.toString
window.Object.prototype.toString = function () {
  if (this instanceof Error)  {
    return "[object Error]";
  }
  else {
    originalObjectToString.apply(this, arguments);
  }
}
```

and after this I got the magical stack traces for the subclasses of `Error`.

And BTW, this is the only way I could find to have the stack traces in EmberJS.

coryvirok:
Thanks for the info, @formigarafa. Does the error subclass have a `.stack` property? If so, this code: https://github.com/rollbar/rollbar.js/blob/master/src/notifier.js#L169 should recognize it as an error object.

If it doesn't have a `.stack` property, where are you seeing the stack traces coming from?

Thanks!

formigarafa:
@coryvirok, the execution only reach that point `if (argT === 'error')` that never happens with subclasses of Error.

[...]

PR #142, opened 2015-09-29
Recognise of objects with stack property as error

formigarafa:
When a Subclass of Error or any other object with a stack is provided with rollbar it is understood as the error argument.
Then the stack is read from it and made available with the respective rollbar item.

This fix what I have seen at #63.


coryvirok:
Ah, now I understand what you were saying. Ya... this looks good and thanks for the PR!

formigarafa:
np. have a good one.

PR #147, opened 2015-10-13
Fix snippet replacement @brianr

coryvirok:


brianr:
Did you mean to delete `docs/faq.md`?

coryvirok:
Yes. It was out of date - It assumed the snippet was using a protocol-less URL when, in fact we hardcode it to https://

brianr:
Got it, thanks.

PR #152, opened 2015-11-16
Don't save shim's onerror when we are building globalnotifier

jondeandres:
We were saving shim's on error on globalnotifier oldOnError. This didn't
have an effect in normal behavior since we ensure to not send the last
error twice, https://github.com/rollbar/rollbar.js/blob/master/src/notifier.js#L652.

When testing rollbar.js in a inspector using `window.onerror()` we are
actually sending the report twice.


coryvirok:
good find

PR #162, opened 2015-12-15
Load full source with crossOrigin

msokk:
Hi!

I noticed that your CloudFront hosted snippet has `Access-Control-Allow-Origin: *` set.
It would be nice to load the full code with crossorigin attribute as then any runtime or syntax errors inside Rollbar.js get properly reported.

Blank string means `anonymous`.
More at https://bugsnag.com/blog/script-error


coryvirok:
Great idea and thanks for the PR. I'll test and merge.

PR #164, opened 2015-12-16
Fixed IE9- bugs and relevant tests

coryvirok:
I removed our custom merge/copy/log polyfills and replaces with
some community ones. Which means we now have some dependencies.

There were a few bugs in how we checked types in IE9- and how
we tested _enqueuePayload() and _log().

Lastly, I fixed a bunch of tests that were failing in IE9 and
verified that all tests now pass in IE9 on Browserstack.

@brianr


src/notifier.js
```diff
@@ -160,7 +151,7 @@ NotifierPrototype._getLogArgs = function(args) {
     } else if (argT === 'date') {
       extraArgs.push(arg);
     } else if (argT === 'error' ||
-               arg.stack ||
+               arg instanceof Error ||
```
brianr: This looks like it will break things if any users were passing in an object with a `stack` property. I'm not sure if anyone is doing this but it seems possible. Any reason not to still support it? If so, we should bump the version number (minor version should be fine, since it's not part of the public api).

src/notifier.js
```diff
@@ -708,15 +699,22 @@ NotifierPrototype.uncaughtError = _wrapNotifierFn(function(message, url, lineNo,
...
 NotifierPrototype.global = _wrapNotifierFn(function(options) {
   options = options || {};
+  var knownOptions = {
+    startTime: options.startTime,
+    maxItems: options.maxItems,
+    itemsPerMinute: options.itemsPerMinute
+  };
 
-  Util.merge(window._globalRollbarOptions, options);
+  extend(true, window._globalRollbarOptions, knownOptions);
```
brianr: What would happen if someone was sending in an option that isn't in `knownOptions`?

src/notifier.js
coryvirok: Yup, I'm planning on a minor bump.

Relying on the stack property is not a great idea since it doesn't provide any guarantees that we will be able to get useful information from it. Also, if someone passes in a plain object with a stack property we were treating it as an error which is probably not what they would want. Also, IE9- doesn't provide a .stack property so we cannot rely on it.

src/notifier.js
coryvirok: We were giving the user the ability to add *any* keys to the global options. This change limits which keys will be put into the global options object based on which ones we support. With this change, if the user passes in an option we don't support, we will silently not put it into the global options object.

brianr: Are there any keys that a user could have put into global options, which would have had an effect?

Based on the docs, it appears that the answer is no, but I want to make sure I'm not missing a way that this could be backwards-incompatible.

src/notifier.js
brianr: Sounds good.

package.json
```diff
@@ -7,6 +7,11 @@
   },
   "license": "MIT",
   "main": "dist/rollbar.umd.nojson.min.js",
+  "dependencies": {
+    "console-polyfill": "0.2.2",
+    "error-stack-parser": "git://github.com/rollbar/error-stack-parser.git",
```
ecneladis: git:// protocol is insecure, please use https:// instead

PR #180, opened 2016-02-10
Ignore errors by body message when a trace is not available

Joshuaweiss:


coryvirok:
Thanks @Joshuaweiss. I added some tests and published v1.8.4.

Issue #107, opened 2015-07-02
Store events to send later if the network isn't available


brianr:
(This is mostly relevant for mobile apps using embedded javascript.)

This feature means roughly: detect when the XHR post in `_processPayload` fails because the network is unavailable, and keep payloads on the queue to be processed later. Periodically retry.

Would pair nicely with #79, but isn't dependent on it.

jonahx:
Fwiw, I've been searching for airbrake alternatives because they lack this feature.  Pretty surprised to see something I consider core to error reporting missing from the clients of the two biggest error tracking SaS companies.

brianr:
@jonahx Thanks for the upvote on this :)

Question--is your situation that you have a web app where clients frequently get disconnected? A mobile app using a webview? Something else?

jonahx:
Yeah, it's a web app often used on mobile ipads, where disconnects are not uncommon.  So I'd missing a ton of errors, and perhaps the most important ones, without this feature.  But my use case aside, XHR requests fail from time to time, even on home connections.  I was originally using a home-rolled email error reporter, 40 lines of code, and it had a retry feature.  I was using just storing the errors in memory, but replacing that with local storage would be trivial.  Just saying, you could guys could probably add this in a few hours to a day of work.

code-constructor:
+1

pimterry:
I've had a run at fixing this: see #186.

coryvirok:
Closing since @pimterry PR was merged.

PR #186, opened 2016-03-18
Retry events that failed to send due to network failures

pimterry:
This is a first run at fixing #107. If a payload fails to be delivered to Rollbar because of something that's likely to be a network failure (a response back without any status code, or a network timeout) then the payload is put back on queue 10 seconds later to be retried.

This is a first run to scope this out and start some discussion. There's a good few debatable points here:

- What should happen with the callback for the event that's being logged? Currently I've opted to be maximally backward compatible: it fires on the first error, exactly as now, and future retries never trigger it, whether they're successful or not.
- The retry delay should presumably be configurable? Should it be disableable?
- Should we eventually stop retrying?
- Should this be some sort of exponential backoff approach instead? (adds a little extra complexity; there's currently no real state between retries)

Regardless, I think this is probably a workable approach in the meantime, so you could potentially treat those as extensions to follow if you like.

There's notably a few more complex approaches to solving this that I'm totally ignoring here (like queuing into LocalStorage and firing later on totally separate page loads, or triggering all waiting previously failed payloads if one of them is successfully delivered). I expect the minimal approach gets most of the benefit of this for most users, with much less effort.

### DIFF
--- a/src/notifier.js
+++ b/src/notifier.js
@@ -43,8 +43,6 @@ function _notifyPayloadAvailable() {
   }
 }
 
-
-
 // Updated by the build process to match package.json
 Notifier.NOTIFIER_VERSION = __NOTIFIER_VERSION__;
 Notifier.DEFAULT_ENDPOINT = __DEFAULT_ENDPOINT__;
@@ -62,6 +60,8 @@ Notifier.LEVELS = {
   error: 3,
   critical: 4
 };
+
+Notifier.RETRY_DELAY = 1000 * 10;
 
 // This is the global queue where all notifiers will put their
 // payloads to be sent to Rollbar.
@@ -1017,7 +1017,7 @@ function _processPayload(payloadObject) {
         payloadObject.callback = function () { };
         setTimeout(function () {
           directlyEnqueuePayload(payloadObject);
-        }, 1000 * 10);
+        }, Notifier.RETRY_DELAY);
       }
 
       return callback(err);
--- a/src/notifier.js
+++ b/src/notifier.js
@@ -16,6 +16,7 @@ var Util = require('./util');
 var xhr = require('./xhr');
 
 var XHR = xhr.XHR;
+var ConnectionError = xhr.ConnectionError;
 var RollbarJSON = null;
 
 function setupJSON(JSON) {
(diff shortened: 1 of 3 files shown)


src/notifier.js
```diff
@@ -1005,6 +1012,14 @@ function _processPayload(url, accessToken, payload, callback) {
   // go ahead and send it.
   XHR.post(url, accessToken, payload, function xhrCallback(err, resp) {
     if (err) {
+      if (err instanceof ConnectionError) {
```
coryvirok: We will want to check to make sure that `ConnectionError` is not `undefined` before doing this `instanceof`

src/notifier.js
```diff
@@ -1005,6 +1012,14 @@ function _processPayload(url, accessToken, payload, callback) {
...
+        // We're calling the callback now with the error, disable the callback for future attempts.
+        payloadObject.callback = function () { };
+        setTimeout(function () {
+          directlyEnqueuePayload(payloadObject);
+        }, 1000 * 10);
```
coryvirok: Let's move this timeout into constant. I think future iterations can implement the exponential backoff by storing `payloadObject._retryAttempts` and using that to calculate the timeout + know when to drop the error altogether.

src/xhr.js
```diff
@@ -8,6 +8,15 @@ function setupJSON(JSON) {
   RollbarJSON = JSON;
 }
 
+function ConnectionError(message) {
```
coryvirok: Cool!

coryvirok:
Thanks for the PR. This is great!

For the discussion points:

### What should happen with the callback for the event that's being logged? Currently I've opted to be maximally backward compatible: it fires on the first error, exactly as now, and future retries never trigger it, whether they're successful or not.

This makes sense to me. I think that if we implement exponential backoff and removal of payloads to be sent after a certain number of attempts we can log a message to the browser about dropping the payload.

### The retry delay should presumably be configurable? Should it be disableable?

I think it should at least be put into a constant and then we can add it to the configuration options exposed via the `global()` method at a later time.

### Should we eventually stop retrying?

Probably but since this is only ever going to last as long as the browser tab is open this should be fine for now.

### Should this be some sort of exponential backoff approach instead? (adds a little extra complexity; there's currently no real state between retries)

Ya but not necessary for this first pass.

pimterry:
Now moved into a constant. I think that's the only change that needed doing to merge this, is that right?

I've also given the built result a quick manual test in my project, seems to work perfectly for me.

coryvirok:
Yup, that was the only thing left. I'm going to test this today and get it merged in.

Thanks again!

Issue #182, opened 2016-03-08
Disable Rollbar.js when adblock/ghostery


fform:
I get a lot of errors from libs though my site when users adblock or browser extension is ripping out javascript. 

Is there an option to just not send events from those users? Does anyone have a solution for this scenario? I get that I could more safely call these libs but this site is at a stage where it's not worth implementing a bunch of re-writing when I could just flip a flag to ignore clients who are intentionally modifying their stack.

piotrd:
One option is a function called `checkIgnore` (https://rollbar.com/docs/filter-or-ignore-errors-from-bots/), which would allow you to ignore certain errors from being sent to Rollbar. Detect if adblocker is active and if so, return `true` from it.

Crisfole:
Hey @fform, @piotrd is right, `checkIgnore` would do the trick. You could also try simply disabling the error reporting when you notice a missing script:

    rollbar.configure({enabled: false});

PR #190, opened 2016-04-13
Adding example and documentation related to adblockers

corps:
@coryvirok @brianr 

This addresses [Issue 182](https://github.com/rollbar/rollbar.js/issues/182) by adding an example an a related doc file.  Let me know of any improvements you'd like to see!

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,13 @@ Rollbar.configure({enabled: false});
 Rollbar.error("This will *not* be reported to Rollbar");
 ```
 
+## Dealing with adblocker / browser extension exceptions
+
+Unfortunately, some very popular browser extensions may modify a user's copy of your website in such a way as
+to break / block it's functionality.  This can result in Rollbar reporting exceptions that are not a direct result
+of your own code.  There are multiple approaches to dealing with this issue, the simplest of which is covered
+ [here](https://github.com/rollbar/rollbar.js/tree/master/docs/extension-exceptions.md)
+
 ## Ignoring specific exception messages
 
 If you want to ignore a specific exception message, say for a third-party browser plugin
--- a/README.md
+++ b/README.md
@@ -108,9 +108,9 @@ Rollbar.error("This will *not* be reported to Rollbar");
 ## Dealing with adblocker / browser extension exceptions
 
 Unfortunately, some very popular browser extensions may modify a user's copy of your website in such a way as
-to break / block it's functionality.  This can result in Rollbar reporting exceptions that are not a direct result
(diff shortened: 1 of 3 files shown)


docs/extension-exceptions.md
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  if (window.addEventListener !== undefined) {
+    window.addEventListener('load', onLoadStartAddCheck, false);
+  } else {
+    window.attachEvent('onload', onLoadStartAddCheck);
+  }
+```
+
+The above approach is likely to work in the majority of cases, *but it is not fullproof*.  Extensions and their
```
brianr: s/fullproof/foolproof/

docs/extension-exceptions.md
```diff
@@ -0,0 +1,77 @@
...
+  };
+</script>
+```
+
+Add an html element with "bait" class names to be removed by adblockers.
+
+```html
+<!-- bait for adblocker like addons.  If they remove this div, we should disable error reporting -->
+<div id="blocker-bait" style="height: 1px; width: 1px; position: absolute; left: -999em; top: -999em"
+     class="ads ad adsbox doubleclick ad-placement carbon-ads"></div>
```
brianr: Clever. What adblockers have you tested this with?

README.md
```diff
@@ -105,6 +105,13 @@ Rollbar.configure({enabled: false});
 Rollbar.error("This will *not* be reported to Rollbar");
 ```
 
+## Dealing with adblocker / browser extension exceptions
```
brianr: Comment regarding placement in this file: can you move it closer to the end, just before the "ios9 webviews" section?

docs/extension-exceptions.md
corps: AdBlock and AdGuard.  Though I suspect a larger amount of ad blockers would "fall" for the bait.  But there really isn't a perfect solution here -- ad blockers are in an arms race to circumvent "adblocker-blockers" and may change behavior in the future :/

But I really think this will work for the two most popular options I gave above.  Anything more becomes an exercise in patience and experimentation.

docs/extension-exceptions.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+Add functions to check for the presence and visibility of our bait div.
+
+```js
+  function disableRollbar() {
+    Rollbar.configure({enabled: false});
```
ezarowny: It's a bummer that the best case scenario is to just disable Rollbar if we have problematic browser extension. I can't think of any other way around this though.

Issue #149, opened 2015-10-20
Is there a reason for the bower package to have 4.1 MB?


jirihelmich:
It should, IMHO, be as small as possible to enable quick build on integration servers and different automated environments. 4.1MB is just huge. Unarchived jquery bower package has 800 kB, for instance.

lparry:
I was just looking into why it's so big myself. in the `release` directory it seems to include every release ever. Not sure if this is an oversight or intentional, but it doesn't seem right to me

corps:
Is this issue outdated?  I just took a peak at the bower install directory for rollbar.js, and the total size was 1.1mb, it seems that the release directory is no longer included.  It's possible to reduce this size a bit further, but the gains will be diminished, and would only involve removing optionally useful things (like various loader versions of the build, src/ directory).

One other possibility that might be more generally useful would be to introduce a separately packaged "rollbarjs-ci" that only packages the bare minimums to execute in the common environments (say, removing the src/, snippets, documentation, etc), but it's not clear if the extra complexity is warranted.  Let me know you guys thoughts.

brianr:
@corps Looking at [bower.json](https://github.com/rollbar/rollbar.js/blob/master/bower.json), it looks like we may really only need to have one file in it, `dist/rollbar.umd.js`. If that's correct, then it seems we could make the package nice and small - down to ~80kb.

corps:
Sure thing.  We should also consider announcing such a change as breaking, however, since any bower consumer who was relying on the presence of source maps, or other specific files, would suddenly break.

brianr:
Hm, good point.

@lparry and @jirihelmich , how does the current package size work for you? Are we still unusually huge or are we into the normal zone? Would removing the `src/` folder and the other files in `dist/` break anything for you?

jirihelmich:
We use just rollbar.min.js from dist folder of that package to reference it into our application.

PR #191, opened 2016-04-13
Removing unneeded files from bower release

corps:
Addressing https://github.com/rollbar/rollbar.js/issues/149

Concern remains over wether this might break some builds that depend on js maps or using .src files (not unheard of, even at my previous company).

Tested to verify this would produce the expected bower install this way:

```bash
> cd rollbar.js
> mv .git .git2 # required, bower will try to checkout tagged version without consulting local files
> bower install ./
> cd bower_components/rollbar
> ls -la 
-rw-r--r--  1 dame  staff  10229  4 13 08:21 CHANGELOG.md
-rw-r--r--  1 dame  staff   1070  4 13 08:21 LICENSE
-rw-r--r--  1 dame  staff  11840  4 13 09:59 README.md
-rw-r--r--  1 dame  staff    498  4 13 12:21 bower.json
-rw-r--r--  1 dame  staff   4043  4 13 08:21 browserstack.browsers.js
-rw-r--r--  1 dame  staff  82695  4 13 08:21 browserstack.browsers.json
-rw-r--r--  1 dame  staff    986  4 13 08:21 defaults.js
drwxr-xr-x  4 dame  staff    136  4 13 08:21 dist
> ls dist/
plugins        rollbar.umd.js
> cd ..
> du -sh ./* 
 12K ./CHANGELOG.md
4.0K ./LICENSE
 12K ./README.md
4.0K ./bower.json
4.0K ./browserstack.browsers.js
 84K ./browserstack.browsers.json
4.0K ./defaults.js
 84K ./dist
```

Alot smaller.


corps:
This has been updated to be a bit more safe about which files are removed based on feedback in the issue (usage of the min file directly).  Basically, the snippet, named-amd, and nojson versions are removed and other files left as is.

coryvirok:
The next release will have a new minor version number and the CHANGELOG will note that the bower package may have breaking changes for users.

coryvirok:
Note: We can also remove the browserstack.* files. These are only used for building/testing of the library and not used by the notifier during customer use.

bower.json
```diff
@@ -5,10 +5,16 @@
     "dist/rollbar.umd.js"
   ],
   "ignore": [
+    "dist/*.snippet*",
```
browniebroke: I was trying to update, but the example is now outdated: https://github.com/rollbar/rollbar.js/tree/master/examples/bower#using-rollbar-with-bower
Should we update the instructions, or keep delivering the snippet?

coryvirok: Thanks for pointing this out. I believe you can use rollbar.min.js from the bower installation but I'm going to re-add the snippet for backwards compatibility.

coryvirok: I just released v1.9.1 which has the snippet file included in the Bower distribution.

bower.json
```diff
@@ -5,10 +5,16 @@
...
+    "dist/*.nojson*",
+    "dist/*.named-amd*",
```
sandstrom: Why is `named-amd` removed? (https://github.com/rollbar/rollbar.js/pull/151)

Or have I misunderstood the context in which it's removed?

sandstrom: friendly ping @coryvirok

coryvirok: Thanks for the ping @sandstrom. I missed it the first time.

The `named-amd` build target is still created. This only removes it from the bower distribution since the UMD target is listed as the `main` JS file.

sandstrom: Sorry if I'm slow, but we're currently using the build from the Bower distribution. For example `path/to/bower/components/rollbar/dist/rollbar.named-amd.nojson.js`

corps: If you're using bower only to perform the download, then pointing your build to a file inside the resulting directory, then all you should need to do is instead switch to using the npm version, executing a build, and pointing to the named-amd output in dist/.

The typical use case for bower involves automatically consuming the "main" entry of a project's bower.json, which in this case differs from you guys consumption.  I would recommend switching to consuming rollbar.js from npm and changing your build appropriately.

sandstrom: Maybe, but if so we're not alone :) Ember-CLI [use this approach](http://ember-cli.com/user-guide/#javascript-assets). Also, looking at [main](https://github.com/bower/spec/blob/master/json.md#main) it does allow other files.

Another option would be to make the named AMD default.

Issue #173, opened 2016-01-07
Some parts of  _rollbarConfig ignored in bower example?


InstanceOfMichael:
```
<head>
  <script>
    _rollbarConfig = {
      accessToken: "ACCESS_TOKEN",
      captureUncaught: true,
      payload: {
         environment: "test"
      }
    };
  </script>
</head>
```

It seems that ``_rollbarConfig.payload.environment`` and any variation of ``_rollbarConfig.*.(person|user)`` are ignored in the bower example and must be manually set via ``Rollbar.config()``

dlsteuer:
I added some additional configuration options to the bower example in #202

PR #202, opened 2016-05-13
added additional options to the bower example, for Issue #173

dlsteuer:
Added some options to verify they work, and added the missing function, so the example works out of the box after adding an ACCES_TOKEN

### DIFF
--- a/examples/bower/index.html
+++ b/examples/bower/index.html
@@ -15,7 +15,12 @@
       };
       
       function breakme() {
-        Rollbar.error("This will be reported to Rollbar");
+        try {
+          var foo = null;
+          foo();
+        } catch (e) {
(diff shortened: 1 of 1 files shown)


examples/bower/index.html
```diff
@@ -5,9 +5,18 @@
...
+          person: {
+            id: 456,
+            username: "foo",
+            email: "foo@example.com"
+          }
         }
       };
+      
+      function breakme() {
+        Rollbar.error("This will be reported to Rollbar");
```
brianr: Would be nice to show an actual Error object being reported here (so that there is a stack trace). e.g.

```js
try {
  var foo = null;
  foo()
} catch (e) {
  Rollbar.error(e);
}
```

Issue #205, opened 2016-06-03
Object.create introduced in version 1.8 doesn't work on ie8


ctrager:
Object.create not supported in IE8.  Was not in rollbar.js 1.7, but is in rollbar.js 1.8
![screen shot 2016-06-03 at 5 24 14 am](https://cloud.githubusercontent.com/assets/61188/15776267/34a529ea-294c-11e6-970d-0026d7d506a9.png)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create

robtarr:
Is there any movement on this? I just picked up a project that is loading v1.9 via https://rubygems.org/gems/rollbar/versions/2.14.0, and it is causing errors in IE8 as well.

Is there fix for this?

PR #206, opened 2016-06-03
Fix to rollbar.js not loading in IE8

dlsteuer:


src/polyfill.js
```diff
@@ -0,0 +1,31 @@
+
+function createPolyfills() {
+    // Object.create not supported in IE8 needs to be polyfilled
```
brianr: Style: 2 spaces per level of indentation

dlsteuer:
This will fix #205

brianr:
As a possible alternative, is there a way we can change the code (i.e. what was added in 1.8) to not need to use Object.create?

dlsteuer:
@brianr here is another way to do it, the browserstack tests are failing for IE8 still but unrelated to this, and this fix works in my testing in an IE8 browser.

src/xhr.js
```diff
@@ -14,7 +14,7 @@ function ConnectionError(message) {
   this.stack = (new Error()).stack;
 }
 
-ConnectionError.prototype = Object.create(Error.prototype);
+ConnectionError.prototype = new Error();
```
coryvirok: Looks like this thread explains some reasons why we shouldn't do this: http://stackoverflow.com/questions/783818/how-do-i-create-a-custom-error-in-javascript

coryvirok:
If you put the logic for the retries into `XHR.post()` you can remove the direct enqueue logic.

PR #207, opened 2016-06-13
Added message when receiving 403 on response

dlsteuer:


src/xhr.js
```diff
@@ -65,10 +65,17 @@ var XHR = {
...
                 // TODO(cory): have the notifier log an internal error on non-200 response codes
+                var jsonResponse = RollbarJSON.parse(request.responseText);
                 if (request.status === 200) {
-                  callback(null, RollbarJSON.parse(request.responseText));
+                  callback(null, jsonResponse);
                 } else if (Util.isType(request.status, 'number') &&
-                            request.status >= 400 && request.status < 600) {
+                  request.status >= 400 && request.status < 600) {
+
+                  if (request.status == 403 && console) {
```
brianr: There are some other places that we use `console` directly; I'm guessing it may work because of this polyfill: https://github.com/rollbar/rollbar.js/blob/master/src/util.js#L1 . Perhaps we should change those all to use a wrapper function that we define in util.js?

src/xhr.js
```diff
@@ -65,10 +65,17 @@ var XHR = {
...
+                    // likely caused by using a server access token, display console message to let
+                    // user know
+                    console.log(jsonResponse.message);
```
brianr: To be consistent with the other log messages, this should start with `"[Rollbar]: "` e.g. https://github.com/rollbar/rollbar.js/blob/master/src/notifier.js#L33

dlsteuer:
I updated the text and switched to console.error like our other messages

brianr:
LGTM.

@coryvirok can you review this as well?

test/notifier.test.js
```diff
@@ -1267,8 +1267,10 @@ describe('Notifier.debug/warn/warning/error/critical()', function() {
       expect(stub.getCall(0).args[0]).to.be.an('object');
       expect(Util.typeName(stub.getCall(0).args[0])).to.equal('error');
 
-      // Shouldn't log internal errors for xhr request failures
-      expect(consoleLogStub.called).to.equal(false);
+      // Shouldn't log internal errors for xhr request failures, we log a message on 403
```
brianr: Can you add a test to cover the 403 case? :)

brianr:
👍

PR #208, opened 2016-06-20
Changed function names from Add to Ad in ad-blocker example

taojonesin:
grammar-nazi_c_254474

Sorry, I used to be an editor for my HS newspaper :/


ezarowny:
:trollface:

brianr:
thanks @taojonesin !

PR #214, opened 2016-08-25
Use sendBeacon when available

tinovyatkin:
Special API to send small payloads, available in Edge, Firefox and Chrome. Reliable and very small, but useful


tinovyatkin:
Sorry, I didn't know regarding single quote convention...

PR #218, opened 2016-09-13
Update captureUnhandledRejections in Quickstart docs

rivkahstandig3636:
A user emailed us to let us know that he thinks this flag should be set to true and not false in our quickstart docs: "I think the simplest way to get someone started quickly and on the right foot with Rollbar would be to make the `captureUnhandledRejections` option set to `true` in your first quick start snippet. That way, the developer using Rollbar will see all errors from the get-go instead of mistakenly assuming their app is having no issues because nothing is showing up on Rollbar, and they can pare back their settings if they're getting overwhelmed by one type of request."

I agreed with him. Hopefully you do too 😸


jondeandres:
:+1:

PR #229, opened 2016-12-22
Retain more of the context on null rejections

holm:
This adds support for getting more context, even when the rejection error itself is null. Without it, there is no information to go on.


rokob:
I think this is a good change, but I'm lacking some context on why in the first place we did not want to enqueue a payload if the reason for rejection was null. @coryvirok any thoughts on this?

coryvirok:
I think Zach, (the person who wrote this code) assumed that a `null` reason was cause for dropping the message. I think this is the better way to do it though. 👍

PR #232, opened 2017-01-17
Serve rollbar.js through CDNJS

jondeandres:


coryvirok:
Very nice. Will the move to CDNJS allow us to serve old rollbar.js versions from CDNJS or will those all need to go through cloudfront?

PR #233, opened 2017-01-19
Remove iOS9 instructions for SSL issues

jondeandres:
It seems that cloudflare support the restrictions by the iOS9 webviews,
so we can remove this section.


rivkahstandig3636:
👍

Issue #236, opened 2017-01-25
Add a way to get the UUID before the callback


ukaaa:
When an error occurs in our app, I would like to log it to Rollbar and then direct the user to a separate error page that displays the UUID. This way they can refer to it when they contact our support team.

Unfortunately, even though the UUID is generated client-side, I can only access the UUID in the callback provided to e.g. `Rollbar.error()`. Would it be possible to get the UUID sooner? Perhaps `NotifierPrototype._log` can return an object that contains contextual information such as the UUID?

ukaaa:
Any thoughts?

brianr:
Hi @ukaaa - great suggestion - we're planning to do this; just figuring out the best way. It's on our list!

ukaaa:
Thanks @brianr, I'm looking forward to that! If I can help in any way please let me know.

lehresman:
@ukaaa Great idea!  Thanks for the suggestion.  I submitted a PR with this functionality.  You'll be able to get the UUID from any of the log functions now.
```javascript
var obj = Rollbar.error('The cake is a lie.');
console.log(obj.uuid);
```

ukaaa:
Thanks @lehresman and sorry for not making a PR myself, I wasn't sure if my suggestion was acceptable to the Rollbar team.

lehresman:
No worries!  Thanks for the suggestion.

lehresman:
Merged.  You'll see this functionality in the next release.

PR #242, opened 2017-02-13
return uuid from log functions, #236

lehresman:
In response to issue #236, this patch makes the log functions return an object containing the generated UUID so it can be used, even before the callback returns.


rokob:
LGTM

Issue #215, opened 2016-08-26
npm package - remove release folder


jirihelmich:
Hi,

I just went through our dependencies and found out that rollbar is our largest (in terms of filesize) dependency - by far!

The tarball for rollbar has over 5 MBs. And most of it is because of the presence of the release folder in the package. That package contains many old released versions.

My question is: Is it OK to have version history in a versioned package? I believe the answer is no. I think it would be really great if you would left just the latest files there.

Could that be done?

Thanks, Jirka

ezarowny:
This is probably a good idea. Let me look into it.

ezarowny:
I'm not quite sure why we need this. I'm going to file an internal ticket to discuss it on Monday.

brianr:
Hey @jirihelmich - thanks for bringing this up. We do keep the old releases around so that we can serve them from our CDN, but I agree with you that they don't need to be in the npm package. We've got this on our todo list.

jirihelmich:
Hey, @brianr, great to hear that! We love your product as it helps us on daily basis. And now, it will be even better! Cheers!

onejgordon:
Big +1 on this, any schedule for removal of unneeded releases in package? According to cost-of-modules, rollbar-browser now sits at 19.76mb. Our largest by a large multiple.

brianr:
Hey everyone, thanks for the attention - consider this bumped back up on our priority list.

PR #243, opened 2017-02-13
ignore the release folder when building the package, #215

lehresman:
I created a `.npmignore` file which is a clone of `.gitignore` but with the `release` folder added.  This will make it so the `release` folder is not included in future builds.

https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,7 @@
+node_modules
+bower_components
+.idea
+vendor/*.min.js
+test/*.bundle.js*
+sauce_connect.log
+release


lehresman:
Fixes #215

rokob:
Looks good to me. Just for my own clarification, the comment by @brianr about keeping the releases around for serving via CDN is not affected by this because it is the `.npmignore` file which only affects the built npm pacakge.

lehresman:
Yes.  The releases will still be in the git repo, just not included when we build a package.

Issue #204, opened 2016-05-26
rollbar module fails to load when using Domino js


lukebayes:
The console-polyfill library does not work properly according to spec as it always attempts to write to the read-only "console" property on window.

I have put together a pull request for them, but in the interim, you might want to fork from the PR branch and inline the fix.

  https://github.com/paulmillr/console-polyfill/pull/25

rokob:
The mentioned PR for console-polyfill has been merged upstream, looks like we just need to ensure that we are using an up to date version of said library.

lehresman:
Thanks @lukebayes.  I've submitted a PR that updates the rollbar.js dependency for console-polyfill to a version that includes your fix.

PR #244, opened 2017-02-13
updated console-polyfill to get the latest fixes

lehresman:
I updated to v0.2.3 of console-polyfill which contains a fix for the problem mentioned in #204.


rokob:
Looks good

PR #245, opened 2017-02-13
remove a remnant from our old usage of TraceKit

### LINKED ISSUE #132: Remove submodule for TraceKit
A client with [Demeteorizer](https://github.com/onmodulus/demeteorizer/) is having trouble. The submodule is being installed as an npm dependency, which isn't legal since it's upper case.

lehresman:
Fixes #132.  We had a legacy reference to TraceKit, but we haven't been using it since v1.3.0


brianr:
@lehresman I think you might also need to remove the `vendor/TraceKit` dir?

rokob:
This looks good to me.

Issue #128, opened 2015-09-02
console output of [object Object] in some browsers is unhelpful


raorao:
problem referenced in [notifier.js](https://github.com/rollbar/rollbar.js/blob/master/src/notifier.js#L579-L580) but I wanted to make an issue for easy linking (as line numbers can change).

Crisfole:
A possible shim for it...if we want to get into the business of shimming lots of different things...http://patik.com/blog/complete-cross-browser-console-log/

lehresman:
I've looked into this console.log replacement, and it looks good for error reporting, but there are two main issues I see:

1.  It is not really a shim, in that it doesn't replace the `console.*` api.  Instead, it provides its own API.  We would need to write a wrapper around `console.*` which calls this.  That's not the end of the world, and could work really well. 
2. The big problem I've seen though, is that the line numbers it reports are inaccurate.  They seem to be off by a few lines, even in the examples provided on their website.  I'm not sure what's up with that, but that seems like a show-stopper, especially since we're trying to provide accurate information to developers.

It would, though, solve the problem of [object Object] being shown on IE browsers.

lehresman:
Another option is to do browser detection and use `JSON.stringify` to output objects to console.log on browsers that would otherwise show `[object Object]`  (i.e.  IE)

rokob:
As to 1, we only use `console.info` to log an object, the other places use `console.error` with expections and strings, so I think we would only need to shim `info` by basically using this `log` code but setting the level to info.

As for 2, I haven't checked completely but it looks like there is an option for including the real line numbers which is false by default.

Looking at how we are using it, I think doing `RollbarJSON.stringify` for some browsers would be okay as this is only for verbose debugging info anyway. The browser detection is not easy though right? We already have the console-polyfill so that it is safe to call methods on the console even if it isn't there, we just need to know for what versions of IE does `console.info(obj)` print `[object Object]`.

lehresman:
I'm hesitant to recommend this `console.log` replacement option.  Reading through the docs, and the nice writeup that was linked above, it's clear this is only intended to be used in development mode.  On older IE browsers (8 and below) it actually loads Firebug Lite from getfirebug.com.  Since this JS notifier will be installed on production websites and not just development mode, that is probably not a great requirement to add.

While I am also hesitant to recommend browser detection, it would be the least intrusive option and would directly address this ticket.  As far as I can tell, the browsers affected are IE 8 and below.  

[...]

PR #249, opened 2017-02-20
log out json for objects in IE8 and below

lehresman:
Addresses #128 by providing a customized console.log/error/info renderer to prevent the dreaded `[object Object]` by rendering JSON instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "license": "MIT",
   "main": "dist/rollbar.umd.nojson.min.js",
   "dependencies": {
-    "console-polyfill": "0.2.3",
+    "console-polyfill": "rollbar/console-polyfill#eab6ff9d2b7597fc2f259baa18100556bdb94dbe",
     "error-stack-parser": "1.3.3",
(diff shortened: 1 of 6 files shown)


rokob:
Looks good, and the tests are nice!

holm:
This seems to have broken console logging in Firefox. 

The use of `apply` with a null `this` is the cause. I think `console.info.apply(null, arguments)`, should be `console.info.apply(console, arguments)`.

rokob:
Looking into this

holm:
We got the error with Firefox 47, since it seems they allowed the use of boundless methods on the console in Firefox 49: https://bugzilla.mozilla.org/show_bug.cgi?id=989619

rokob:
It is an easy fix to call `console.log.apply(console, arguments)` as long as that doesn't break anywhere else. But that should be fine. Bound should work equally well if the unbound version does.

holm:
Created a PR for this change here: https://github.com/rollbar/rollbar.js/pull/252

PR #252, opened 2017-03-16
Invoke console methods with right this-context

holm:
Changes as discussed in https://github.com/rollbar/rollbar.js/pull/249


rokob:
LGTM, I just fixed the issue with eslint on master so this should be good to go.

PR #254, opened 2017-03-23
Universal

rokob:
Well here is the PR, I will try to go through an annotate the important bits, and this isn't quite working completely, but it is close enough to get looked at.


src/rollbar.js
```diff
@@ -0,0 +1,82 @@
...
+      callback = item.callback;
+      delete item.callback;
+    }
+    item.level = item.level || defaultLevel;
+    this.notifier.log(item, callback);
+  }, this)();
+};
+
+Rollbar.prototype._defaultLogLevel = function() {
+  return this.options.logLevel || 'debug';
```
coryvirok: We have the default in package.json: https://github.com/rollbar/rollbar.js/pull/254/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L104

src/server/parser.js
```diff
@@ -0,0 +1,317 @@
+/*jslint devel: true, nomen: true, plusplus: true, regexp: true, indent: 2, maxlen: 100 */
+
+"use strict";
+
+var logger = require('./logger');
+var async = require('async');
```
coryvirok: Is this going to cause problems with newer JS interpreters since "async" is a reserved keyword?

src/server/rollbar.js
```diff
@@ -0,0 +1,233 @@
...
+Rollbar.prototype.global = function(options) {
+  this.client.global(options);
+};
+
+Rollbar.prototype.configure = function(options) {
+  this.options = _.extend(true, {}, this.options, options);
+  this.client.configure(options);
+};
+
+Rollbar.prototype.initialize = function(options, client) {
```
coryvirok: Can we just move this into the constructor? When would `initialize()` be called externally?

src/server/rollbar.js
```diff
@@ -0,0 +1,233 @@
...
+
+  this.client = client || new Client(this.context, this.options);
+
+  addTransformsToNotifier(this.client.notifier);
+  addPredicatesToQueue(this.client.queue);
+
+  if (this.options.handleUncaughtExceptions) {
+    this.handleUncaughtExceptions();
+  }
+  if (this.optiions.handleUnhandledRejections) {
```
coryvirok: typo

src/server/rollbar.js
```diff
@@ -0,0 +1,233 @@
...
+  this.options = _.extend(true, {}, this.options, options);
+
+  options.environment = options.environment || process.env.NODE_ENV || 'unspecified';
+
+  this.client = client || new Client(this.context, this.options);
+
+  addTransformsToNotifier(this.client.notifier);
+  addPredicatesToQueue(this.client.queue);
+
+  if (this.options.handleUncaughtExceptions) {
```
coryvirok: Looks like this will happen for each instance of the notifier which doesn't seem correct.

src/server/rollbar.js
```diff
@@ -0,0 +1,233 @@
...
+
+  if (err instanceof Error) {
+    return this.error(err, request, cb);
+  }
+  return this.error('Error: ' + err, request, cb);
+};
+
+/** DEPRECATED **/
+
+Rollbar.prototype.reportMessage = function(message, level, request, callback) {
```
coryvirok: Let's also add in a "deprecated" log warning.

src/server/rollbar.js
```diff
@@ -0,0 +1,233 @@
...
+  delete this.options.exitOnUncaughtException;
+
+  process.on('uncaughtException', function(err) {
+    this.error(err, function(err) {
+      if (err) {
+        logger.error('Encountered error while handling an uncaught exception.');
+        logger.error(err);
+      }
+
+      if (exitOnUncaught) {
```
coryvirok: Should we call `wait()` before exiting?

src/server/transforms.js
```diff
@@ -0,0 +1,246 @@
...
+var url = require('url');
+var _ = require('../utility');
+
+var defaultSettings = {
+  host: os.hostname(),
+  environment: 'development',
+  framework: 'node-js',
+  showReportedMessageTraces: false,
+  notifier: {
+    name: 'node_rollbar',
```
coryvirok: This should probably be "rollbar.js" but it might require some changes in the pipeline.

src/server/transport.js
```diff
@@ -0,0 +1,157 @@
...
+ */
+
+function get(accessToken, options, params, callback, transportFactory) {
+  if (!callback || !_.isFunction(callback)) {
+    callback = function() {};
+  }
+  options = options || {};
+  _.addParamsAndAccessTokenToPath(accessToken, options, params);
+  options.headers = _headers(accessToken, options);
+  var t;
```
coryvirok: Style: move var declaration to the top of the function.

src/server/parser.js
rokob: It turns out that `async` alone is not a new reserved word, the phrase `async function` is reserved, probably because `async` is so commonly used as a variable. `await` is a new reserved word.

src/apiUtility.js
```diff
@@ -0,0 +1,92 @@
+var _ = require('./utility');
+
+function buildPayload(accessToken, data, jsonBackup) {
+  if (_.isType(data.context, 'object')) {
```
brianr: I think this is because `data.context` is required to be a string per our API spec. I think this is the related issue from node_rollbar: https://github.com/rollbar/node_rollbar/issues/73

src/browser/rollbar.js
```diff
@@ -0,0 +1,269 @@
...
+    item = this._createItem([message, reason, context]);
+  } else {
+    item = this._createItem([message, context]);
+    item.stackInfo = _.makeUnhandledStackInfo(
+      message,
+      '',
+      0,
+      0,
+      null,
+      'unhandledrejection',
```
rokob: Those are actually different parameters, `unhandledrejection` here corresponds to `onerror` above. It is the mode value that gets put into the stack trace.

src/browser/rollbar.js
```diff
@@ -0,0 +1,269 @@
...
+
+function Rollbar(options, client) {
+  this.options = _.extend(true, defaultOptions, options);
+  var context = 'browser';
+  this.client = client || new Client(context, this.options);
+  this.init(this.client);
+}
+
+Rollbar.prototype.global = function(options) {
+  this.client.global(options);
```
rokob: This does configure the global state stored on the Rollbar object instead of on the prototype instances.

coryvirok: Ah, I see that this is configuring the RateLimiter singleton. Do you think we should rename this method or move the configuration for the global rate limiter to no longer be accessible via each Rollbar instance?

src/server/parser.js
coryvirok: aha, thanks!

src/queue.js
```diff
@@ -0,0 +1,217 @@
...
+
+/*
+ * addItem - Send an item to the Rollbar API if all of the predicates are satisfied
+ *
+ * @param item - The payload to send to the backend
+ * @param callback - function(error, repsonse) which will be called with the response from the API
+ *  in the case of a success, otherwise response will be null and error will have a value. If both
+ *  error and response are null then the item was stopped by a predicate which did not consider this
+ *  to be an error condition, but nonetheless did not send the item to the API.
+ */
```
rokob: I thought I was going to need `force` to allow some requests to be sent even if we are trying to drain the queue with wait, but that didn't turn out to be needed so I can get rid of it for the time being.

src/server/rollbar.js
```diff
@@ -30,6 +30,15 @@ function Rollbar(options, client) {
   }
 }
 
+Rollbar.instance = null;
```
coryvirok: Can we add static methods to the Rollbar Object and proxy those calls to the singleton?

E.g.
```js
Rollbar.debug(...) {
    if (Rollbar.instance) {
        return Rollbar.instance.debug(...)
    } else {
        throw Exception('Rollbar not initialized');
    }
}
```

rokob: Yeah I was actually just getting around to doing that because the way I did it with instance doesn't quite work that well.

src/server/rollbar.js
```diff
@@ -30,33 +30,59 @@ function Rollbar(options, client) {
   }
 }
 
-Rollbar.instance = null;
-Rollbar.init = function(options) {
-  if (Rollbar.instance) {
-    return Rollbar.instance;
+_instance = null;
```
coryvirok: Missing `var`

examples/no-conflict/test.html
```diff
@@ -0,0 +1,55 @@
...
+            return true;
+          }
+        } catch (e) {
+          console.error(e);
+        }
+        return false;
+      }
+
+      var token = "POST_CLIENT_ITEM_TOKEN";
+      token = "12c99de67a444c229fca100e0967486f";
```
coryvirok: Is this supposed to be here?

examples/no-conflict/tool.js
```diff
@@ -0,0 +1,11 @@
+import Rollbar from 'rollbar';
+
+const rollbar = new Rollbar({
+  accessToken: '59c9582c45d945f0ad2b7c3dc1aa9813',
```
coryvirok: Should this access token be here?

rokob:
Okay, this is done, kind of hard to review at this point, but ready to merge.

rokob:
Merging

Issue #255, opened 2017-03-24
default `async` option opposite of README


awinograd:
The docs (https://github.com/rollbar/rollbar.js/tree/b04781118f5882095aaa2d5ef3b3cf068fe4372e#synchronous-option) say that by default, including the Rollbar snippet (shim) should load the full Rollbar source asynchronously. However, the opposite appears to be true due to the async option being inverted twice.

If no async option is passed in _rollbarConfig, then loadFull will be called with the third argument sent to `!_rollbarConfig.async => !undefined => true`

https://github.com/rollbar/rollbar.js/blob/b04781118f5882095aaa2d5ef3b3cf068fe4372e/src/bundles/rollbar.snippet.js#L12

This means, 'immediate => true` at:

https://github.com/rollbar/rollbar.js/blob/b04781118f5882095aaa2d5ef3b3cf068fe4372e/src/shim.js#L120

And the script's `async` tag will be set to `!immediate => !true` here:

https://github.com/rollbar/rollbar.js/blob/b04781118f5882095aaa2d5ef3b3cf068fe4372e/src/shim.js#L160

If i'm right, the fix should be an easy one character change, but maybe I'm missing something?

PR #256, opened 2017-03-24
Fix/improve async logic for full script load

awinograd:
* Fix default behavior to match README
* Add defer attribute to script tag for supported browsers

Fixes https://github.com/rollbar/rollbar.js/issues/255

### DIFF
--- a/src/bundles/rollbar.snippet.js
+++ b/src/bundles/rollbar.snippet.js
@@ -9,4 +9,4 @@ _rollbarConfig.rollbarJsUrl = _rollbarConfig.rollbarJsUrl || __DEFAULT_ROLLBARJS
 var shim = RollbarShim.init(window, _rollbarConfig);
 var callback = snippetCallback(shim, _rollbarConfig);
 
-shim.loadFull(window, document, !!_rollbarConfig.async, _rollbarConfig, callback);
+shim.loadFull(window, document, _rollbarConfig.async === undefined || _rollbarConfig.async, _rollbarConfig, callback);
--- a/src/bundles/rollbar.snippet.js
+++ b/src/bundles/rollbar.snippet.js
@@ -9,4 +9,4 @@ _rollbarConfig.rollbarJsUrl = _rollbarConfig.rollbarJsUrl || __DEFAULT_ROLLBARJS
 var shim = RollbarShim.init(window, _rollbarConfig);
 var callback = snippetCallback(shim, _rollbarConfig);
 
-shim.loadFull(window, document, _rollbarConfig.async === undefined || _rollbarConfig.async, _rollbarConfig, callback);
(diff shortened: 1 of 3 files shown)


rokob:
So the README says the default is to set async to true when async is undefined, there are two problems with this change.

1. `!!undefined === !!false === false` whereas we want some function where `f(undefined) === f(true) === true` and `f(false) === false`.
2. Currently, if `_rollbarConfig.async === true` then we call loadFull with immediate equal to `!true === false`. Then we set async to `!immediate === !false === true`. So the current double negation is correct, this change would do the opposite.

I noticed this recently as well, and I think it wasn't obvious before because Chrome loads it async regardless of the presence of the async parameter. The only way I see to do this is:

```
_rollbarConfig.async = _rollbarConfig.async === undefined || _rollbarConfig.async;
```

awinograd:
Ahh, yes you're right. I fixed the default case but broke it when the async option is set. Will update my PR

awinograd:
@rokob, happy monday! Free to take another look?

src/bundles/rollbar.snippet.js
```diff
@@ -9,4 +9,4 @@ _rollbarConfig.rollbarJsUrl = _rollbarConfig.rollbarJsUrl || __DEFAULT_ROLLBARJS
 var shim = RollbarShim.init(window, _rollbarConfig);
 var callback = snippetCallback(shim, _rollbarConfig);
 
-shim.loadFull(window, document, !_rollbarConfig.async, _rollbarConfig, callback);
+shim.loadFull(window, document, _rollbarConfig.async === undefined || _rollbarConfig.async, _rollbarConfig, callback);
```
rokob: This should still be negated
```
!(_rollbarConfig.async === undefined || _rollbarConfig.async)
```

awinograd: Just fixed in 8476b82.

Sorry about the sloppy boolean logic!

awinograd: What would your thoughts be on changing the `immediate` variable to `async` instead? Might be simpler to reason about and would shave off a couple of bites of `!` characters

rokob:
LGTM

awinograd:
Thanks for catching my numerous mistakes and merging!

rmacklin:
@rokob It seems that as part of the refactor in https://github.com/rollbar/rollbar.js/pull/254/files#diff-ec822a34e902804f64cc4f4d376aabd9L161, setting the script `defer` attribute may have been lost. Was that intentional? If so, can you explain why, and should we remove the mention of `defer` from the README: https://github.com/rollbar/rollbar.js/blob/16b07e2aac684926d2ab5d8e081916311d30d62b/README.md#L283

PR #264, opened 2017-04-24
Bug: Leaked sensitive data when stringifying custom data

jbrumwell:
When custom is used as a message it is currently sent unscrubbed. This fix ensures that custom is scrubbed when applied as the body message.

This change does require an extra call to `_scrub` it could be refactored in the following way if desired;

```
var payloadData = {
  ...
  body: { extra : custom },
  ...
};

// continue and scrub payload.data

payload.data.body = this._buildBody(message, stackInfo, payload.data.body.extra);

return payload;
```

This would avoid having to call `_scrub` twice.

### DIFF
--- a/src/notifier.js
+++ b/src/notifier.js
@@ -341,6 +341,11 @@ NotifierPrototype._buildPayload = function(ts, level, message, stackInfo, custom
 
 NotifierPrototype._buildBody = function(message, stackInfo, custom) {
   var body;
+
+  if (custom) {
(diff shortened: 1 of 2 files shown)


brianr:
Thanks @jbrumwell !

I think this issue might be fixed in the `universal` branch (#254)... @rokob can you comment?

rokob:
It does now! Thanks for this, I also missed this when doing the universal update.

rokob:
Fixed in the latest release

PR #276, opened 2017-05-17
Remove examples and coverage from published package

akshayp:
These don't add a lot of value inside of a published tarball. This should shave ~1.5Mb from the installed package


rokob:
👍

Issue #282, opened 2017-05-23
handle logging args identically on server and client


modosc:
first, thanks for the work on this!

unfortunately it's still not possible to use this in a universal app without being _really_ careful with what you're doing.

for instance, i often use a patten like this inside of a react component:
```
Rollbar.error("Some useful context message", e, { some: custom, data: here })
```

this works fine on the client but not on the server side. 

is there a reason that the [server version of _createItem](https://github.com/rollbar/rollbar.js/blob/master/src/server/rollbar.js#L344)  doesn't use the same logic as the [client](https://github.com/rollbar/rollbar.js/blob/master/src/browser/rollbar.js#L217)? 

if it's not intentional i could get a pr together to try and unify, just wanted to make sure i wasn't missing something.

modosc:
i'm actually working on this now.

modosc:
will you be able to roll a new release soon from `master`?

coryvirok:
Thanks for the work on this, @modosc! Can you open a PR so we can review?

brianr:
@coryvirok I think that was PR #299

modosc:
yeah, pretty sure this is [merged already](https://github.com/rollbar/rollbar.js/commit/f67efaefd4c3f7a686f649e534acfc21a3ec2e7d)?

coryvirok:
Woo! Closing.

PR #286, opened 2017-05-24
[#282] handle unordered params on the server side

modosc:
this ports the client param handling over to the server side. the only real difference is attempting to identify a `request` object. the first object found that contains at least one of the optional `request` keys is considered to be a request (this may result in false positives - we could require more than one? a minimal subset? ). otherwise the logic is identical.

i added tests to verify all the old parameter ordering still worked, and i added a single test for unordered params. i couldn't figure out how to metaprogram tests in `vows`, otherwise i would have tested for every combination.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ export class AppModule { }
 
 ### Backbone.js
 
-Rollbar.js supports Backbone.js with no additional configuration required. 
+Rollbar.js supports Backbone.js with no additional configuration required.
 
 ## UMD / Browserify / Requirejs / Webpack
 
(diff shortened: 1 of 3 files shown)


rokob:
I like the idea, I toyed with this at some point and went the easier route to specifying the order for the server. Personally, I would like like to go to a fixed parameter order on both sides, but that isn't going to happen so this is probably the way to go. I will take a look.

modosc:
sorry to bug - curious if you've had a chance to look over? this change will allow me to remove a ton of code in our codebase so i'm pretty anxious to get it merged (or fixed if necessary). thanks!

rokob:
Okay LGTM, I am going to merge this, then I will probably pull this param parsing code into a common utility because now it is basically the same thing on browser and server.

modosc:
> Okay LGTM, I am going to merge this, then I will probably pull this param parsing code into a common utility because now it is basically the same thing on browser and server.

that was my though too but there is the subtle difference with the request object. i couldn't figure out a way to make it generic but still handle that gracefully.

anyway, thanks! looking forward to getting this released.

PR #288, opened 2017-05-26
Do not fail if the request object does not have a headers key (update request-ip)

finnp:
Hey ☺️,

with the latest version having a request object without a headers key fails (even though that's allowed according to the docs). This can be solved by just updating `request-ip` to the latest version. I also added a new test that will check that using an empty request object will not fail.

Thank you for your great product by the way, it's very helpful for us at greenkeeper ✨


rokob:
LGTM, I made that change to request-ip and I forgot to update the package even though that is why I removed the code that used to do the check on our side.

PR #291, opened 2017-05-30
Fixes #290 handle objects without Object in prototype chain

### LINKED ISSUE #290: TypeError: undefined is not a function in obj.hasOwnProperty(k)
querystring objects seem to break rollbar. It may have something to do with https://github.com/nodejs/node/pull/6289

This is the property that is being scrubbed that is causing the failure:

```
  GET: { color: '999' },
```

Exception: 

```
/Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/src/utility.js:129
      if (obj.hasOwnProperty(k)) {
              ^

TypeError: undefined is not a function
    at traverse (/Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/src/utility.js:129:15)
    at scrubber (/Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/src/utility.js:431:16)
    at traverse (/Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/src/utility.js:142:14)
    at scrubber (/Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/src/utility.js:431:16)
    at traverse (/Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/src/utility.js:142:14)
    at Object.scrub (/Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/src/utility.js:439:3)
    at Array.scrubPayload (/Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/src/server/transforms.js:137:5)
    at cb (/Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/src/notifier.js:109:31)
    at Array.addRequestData (/Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/src/server/transforms.js:129:3)
    at cb (/Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/src/notifier.js:109:31)
    at cb (/Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/src/server/transforms.js:78:5)
    at /Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/node_modules/async/lib/async.js:49:16
    at /Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/node_modules/async/lib/async.js:263:32
    at /Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/node_modules/async/lib/async.js:40:16
    at /Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/src/server/transforms.js:162:14
    at /Users/aaronjensen/Source/my-app/functions/node_modules/rollbar/src/server/parser.js:289:12
```

rokob:
Sometimes people use `Object.create(null)` which means that Object does not end up in the prototype chain of said created objects. This means that hasOwnProperty is not a method on that object so calling `obj.hasOwnProperty(prop)` will break. To get around this, we can use `Object.prototype.hasOwnProperty.call(obj, prop)`. As far as I can tell this is safe at least as far back as IE 6 and all version of node. It is only relevant when we don't really know how the object was constructed, so it isn't necessary to use this universally. There is a miniscule performance hit and a bit of a readability issue with it, so prefer not to use this style when we create the object ourselves.

### DIFF
--- a/src/server/parser.js
+++ b/src/server/parser.js
@@ -44,7 +44,7 @@ function getMultipleErrors(errors) {
   errArray = [];
 
   for (key in errors) {
-    if (errors.hasOwnProperty(key)) {
+    if (Object.prototype.hasOwnProperty.call(errors, key)) {
       errArray.push(errors[key]);
     }
   }
--- a/src/browser/rollbar.js
+++ b/src/browser/rollbar.js
@@ -176,9 +176,11 @@ Rollbar.prototype.wrap = function(f, context) {
 
       f._wrapped._isWrap = true;
 
-      for (var prop in f) {
-        if (f.hasOwnProperty(prop)) {
-          f._wrapped[prop] = f[prop];
+      if (f.hasOwnProperty) {
+        for (var prop in f) {
+          if (f.hasOwnProperty(prop)) {
+            f._wrapped[prop] = f[prop];
+          }
         }
       }
     }
--- a/src/browser/shim.js
+++ b/src/browser/shim.js
@@ -148,9 +148,11 @@ Shim.prototype.wrap = function(f, context) {
 
       f._wrapped._isWrap = true;
 
-      for (var prop in f) {
-        if (f.hasOwnProperty(prop)) {
-          f._wrapped[prop] = f[prop];
+      if (f.hasOwnProperty) {
+        for (var prop in f) {
+          if (f.hasOwnProperty(prop)) {
+            f._wrapped[prop] = f[prop];
+          }
         }
       }
     }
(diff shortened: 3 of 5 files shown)


src/server/parser.js
```diff
@@ -44,7 +44,7 @@ function getMultipleErrors(errors) {
   errArray = [];
 
   for (key in errors) {
-    if (errors.hasOwnProperty(key)) {
+    if (Object.prototype.hasOwnProperty.call(errors, key)) {
```
coryvirok: Can you add a comment describing why this is necessary? I can see someone coming along and "fixing" this to use `errors.hasOwnPropery(key)` if they didn't know better.

PR #292, opened 2017-05-31
Actually honor NODE_ENV

djanowski:
Without this change, environment always reported as development.


rokob:
Hah oh man that's terrible, my bad. 👍

brianr:
Wouldn't this change mean that you can't specify the environment name explicitly? i.e. what if you want NODE_ENV to be a different value from the rollbar environment name?

rokob:
No it gets picked up via

`this.options = _.extend(true, {}, Rollbar.defaultOptions, options);`

djanowski:
Thank you!

PR #293, opened 2017-05-31
subclasses of errors were not treated as errors before

rokob:
this fixes it as long as the prototype chain is correct


jelder:
Would be great to have this shipped soon.

PR #296, opened 2017-06-01
Browser singleton

rokob:
We added a singleton helper to the server side, it is something people have been confused about and asked for on the browser side. This is what it would look like.


coryvirok:
Cool. This approach LGTM.

rokob:
There isn't really a downside to merging this beyond adding some API we may not want to support in the future. This makes rollbar.js compatible with basically all pre-2.0 usage.

Issue #298, opened 2017-06-01
server-side code errors out when given a message and an error


modosc:
on the client `Rollbar.error("a message", e)` will silently drop the message if `e` is an error. 

on the server this doesn't happen which results in this error:
```
API error: Invalid format. data.body must have either a 'trace', 'trace_chain', 'message' or 'crash_report' property, (but not more than one).
```

i can make a pr to fix this on the server, but i'm curious if the client side behavior is really the desired behavior?

modosc:
pr: #299

modosc:
ping?

davidgwking:
I'm also encountering this issue. Thanks for the fix, @modosc. Looking forward to response from the team at rollbar.

coryvirok:
Thanks for surfacing this. I added a suggestion for how to do this in the PR.

sarkistlt:
Hi, any updates here? still have this issue

rokob:
I'll get this out today

sarkistlt:
ok, cool

PR #299, opened 2017-06-01
[#298] handle message and error on server side

modosc:


src/server/transforms.js
```diff
@@ -71,6 +71,9 @@ function buildErrorData(item, options, callback) {
   item.data.body = item.data.body || {};
   item.data.body.trace_chain = chain;
 
+  // clear out any existing message
+  delete item.data.body.message
```
coryvirok: I'd like to not lose this information. I think the preferred way to handle this is to implement what the browser transforms do:

From https://github.com/rollbar/rollbar.js/blob/master/src/browser/transforms.js#L99
```js
function addBody(item, options, callback) {
  if (item.stackInfo) {
    addBodyTrace(item, options, callback);
  } else {
    addBodyMessage(item, options, callback);
  }
}
```

I would refactor the server-side to remove the `addMessageData()` transform and add the `addBody()` transform in its place. That code *may* need to be tweaked to remove the item.message in the case we have an error as well. In this case, we should set the `error.description = message`. @rokob can provide more details.

coryvirok:
Thanks for this PR! I think there's a more preferred way to handle this case. See my comment.

modosc:
thans @coryvirok - that's way more detailed of a change than i have time to handle right now. hopefully someone else can pick this up asap?

coryvirok:
No prob. We will implement the fix once @rokob is back in town on Monday.

sarkistlt:
Hi, the same issue, is it already fixed?

fijimunkii:
Hi I am experiencing this same issue. Any progress?

PR #311, opened 2017-06-12
add back verbose logging

rokob:
As far as I can tell, all the verbose option used to do was log to console.error when we are about to log an exception, so this adds that functionality back

### DIFF
--- a/src/queue.js
+++ b/src/queue.js
@@ -216,14 +216,20 @@ Queue.prototype._dequeuePendingRequest = function(item) {
 };
 
 Queue.prototype._maybeLog = function(item) {
-  if (this.logger && this.options.verbose) {
-    var message = _.get(item, 'data.body.trace.exception.message');
-    if (!message) {
-      message = _.get(item, 'data.body.trace_chain.0.exception.message');
-    }
-    if (message) {
-      this.logger.error(message);
-    }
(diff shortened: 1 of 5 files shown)


src/queue.js
```diff
@@ -212,4 +215,16 @@ Queue.prototype._dequeuePendingRequest = function(item) {
   }
 };
 
+Queue.prototype._maybeLog = function(item) {
+  if (this.logger && this.options.verbose) {
+    var message = _.get(item, 'data.body.trace.exception.message');
```
coryvirok: Should this also handle logging messages which are not part of a trace/trace_chain?

rokob: As far as I can tell, https://github.com/rollbar/rollbar.js/blob/7717f62eecedb3f5e7400b633ab8070ed2e45b6b/src/notifier.js#L577 is the only use of the verbose option in the old rollbar.js this basically matches that. I am not sure exactly what we want to log in verbose mode. There were also a lot of info logging to the console going on before which we have essentially removed. I have to dig in to see how that was configured, but it did not use the verbose option.

coryvirok: Gotcha. If it's easy, I'd like to add verbose logging for messages as well as exceptions. Basically, anytime someone calls one of the top-level API methods which result in a POST.

PR #312, opened 2017-06-14
Unhandled rejections [Fixes #307]

### LINKED ISSUE #307: `unhandledrejection` disregards the `reason`, can't get proper stack trace in `userTransform`
Hi,

Test case:
https://jsfiddle.net/etu4g6qc/5/

I'm having an issue with debugging an `unhandledrejection` of a Promise that I"m not handling.  
I"m currently logging such errors to a logging service, and what I see is `[object Object]` and 
a short stack trace from Bluebird which doesn't help at all.

After hours of debugging every piece of code I could, I found this:  
https://github.com/rollbar/rollbar.js/blob/master/dist/rollbar.umd.js#L209
```
Rollbar.prototype.handleUnhandledRejection = function(reason, promise) {
          ...
	  if (_.isError(reason)) {
	    item = this._createItem([message, reason, context]);
	  } else {
	    item = this._createItem([message, context]);
```

`reason` here came from `event.detail` (rejectionHandler at https://github.com/rollbar/rollbar.js/blob/master/dist/rollbar.umd.js#L2653) and it's just an Object, not an Error instance, so the `item` that's created disregards `reason`.

When debugging in a real world test case I saw that `reason` contained all of the information I needed (the actual object sent from the server as an error response).

## Issues:
- The `payload` that's passed to `transform` (from `UserTransform`) does not contain `reason`, so I can't log essential details to the logging service.
- The rejection can be an Object, in which case `body.trace.exception.message` is `[object Object]`

Would appreciate your help.

rokob:
Rejecting with a non-error reason results in `reason.message` or `String(reason)` being set as the message to log. This gets rids of structure that the reason object might have had, so include the reason again to be picked up in custom data. I am not sure I would call it an anti-pattern per se to use non-errors as reasons for rejecting promises, but it is highly non-standard.

### DIFF
--- a/src/browser/rollbar.js
+++ b/src/browser/rollbar.js
@@ -123,7 +123,7 @@ Rollbar.prototype.handleUnhandledRejection = function(reason, promise) {
   if (_.isError(reason)) {
     item = this._createItem([message, reason, context]);
   } else {
-    item = this._createItem([message, context]);
+    item = this._createItem([message, reason, context]);
     item.stackInfo = _.makeUnhandledStackInfo(
       message,
       '',
@@ -138,7 +138,7 @@ Rollbar.prototype.handleUnhandledRejection = function(reason, promise) {
   item.level = this.options.uncaughtErrorLevel;
   item._isUncaught = true;
   item._originalArgs = item._originalArgs || [];
-  item._originalArgs.push(reason, promise);
+  item._originalArgs.push(promise);
   this.client.log(item);
 };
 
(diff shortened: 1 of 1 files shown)


coryvirok:
It is non-standard but it's best to include more information than less especially when JS is so fre-form that you can throw non-exceptions/errors.

coryvirok:
Tagging issue #307

alexilyaev:
Much appreciated!

alexilyaev:
Can we publish this to npm?

PR #318, opened 2017-06-20
add back lastError on the rollbar instance

rokob:
`lastError` used to be a property that was stored on the notifier instance, this was used for a very simple form of deduping of errors; add this functionality back.

### DIFF
--- a/src/browser/rollbar.js
+++ b/src/browser/rollbar.js
@@ -24,7 +24,6 @@ function Rollbar(options, client) {
   if (this.options.captureUnhandledRejections) {
     globals.captureUnhandledRejections(window, this);
   }
-  this.lastError = null;
 }
 
 var _instance = null;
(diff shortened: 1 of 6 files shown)


rokob:
I admit I don't like the repetition here. Part of the problem is how things are currently constructed, another problem is that I can't do property level delegation in JS. One alternative would be to make `lastError` a function instead of a property. Then I could push things down to the shared client, do the deduping there and just expose the lastError function from that shared client as a method on the browser/server rollbar instances.

src/browser/rollbar.js
```diff
@@ -371,6 +399,14 @@ Rollbar.prototype._createItem = function(args) {
   return item;
 };
 
+Rollbar.prototype._sameAsLastError = function(item) {
+  if (this.lastError === item.err && this.lastError) {
```
coryvirok: Should these two conditions be swapped? I imagine the second is more common than the first.

src/browser/rollbar.js
```diff
@@ -87,6 +91,9 @@ Rollbar.log = function() {
 Rollbar.prototype.debug = function() {
   var item = this._createItem(arguments);
   var uuid = item.uuid;
+  if (this._sameAsLastError(item)) {
```
coryvirok: If we transition to having all of the `debug/info/warn/...` methods always call `this.client.log(item, level)` we can abstract this code into the `client.log()` method.

src/browser/rollbar.js
rokob: Yeah, either way for the purposes of this, but yeah the second is probably more common

src/browser/rollbar.js
rokob: Even if we did, the client would then have to track the lastError, but then lastError couldn't be a property on the top level instance, unless we returned lastError from `client.log` or converted lastError to a function

PR #319, opened 2017-06-20
add a blacklist option to the browser side [Fixes #283]

### LINKED ISSUE #283: `hostBlackList` option
This would work similar to the existing `hostWhiteList` option, except that if _any_ frame in the stack trace matches an entry in the list, the error is ignored.

This would be useful for suppressing errors that come from third party code that you don't maintain.

rokob:
We have a hostWhiteList option, this implements the alternative blacklist which allows you to ignore errors which have any frame in their backtrace with a "filename" that matches a regex in this list.

### DIFF
--- a/src/browser/predicates.js
+++ b/src/browser/predicates.js
@@ -32,16 +32,31 @@ function userCheckIgnore(item, settings) {
   return true;
 }
 
+function urlIsBlacklisted(item, settings) {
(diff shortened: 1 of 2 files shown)


rokob:
This won't link with #283 for some reason

rokob:
Those codacy checks are not legit

Issue #320, opened 2017-06-21
Node Server - UnhandledExceptions -> also log to console


gerwinbrunner:
I'm on node and I use this configuration snippet for rollbar:
```
const rollbar = Rollbar.init({
  accessToken: rollbarServerAccessToken,
  handleUncaughtExceptions: true,
  exitOnUncaughtException: true,
  environment: Meteor.settings.public.env,
});

```

I created a test error, by calling `throw new Error('!!! JUHUHUHU !!! CALLED');`
The error gets reported to rollbar and the server stops 👍 

The problem now is, that in the log I do not have any information at all about the issue.
How can I let rollbar handle the exception but also have a regular console.log or console.error statement?

rokob:
There is an option, `verbose`, that should also log to the console. Try adding `verbose: true` to your configuration.

gerwinbrunner:
Hi @rokob did that already. But the option seems to only work on the client, not on the server.
I use rollbar version `^2.0.4`

rokob:
We use the [debug](https://www.npmjs.com/package/debug) package on the node side which requires you to use an environment variable to enable the log output. For example,

```
DEBUG=Rollbar:* node app.js
```

rokob:
That should be documented somewhere and I don't think it is, I will fix that.

gerwinbrunner:
Ok, I tried it...

I now get the folloging output:
```
I20170622-19:07:24.729(2)? PRE - JUHUHUHU C exception
I20170622-19:07:25.687(2)? Thu, 22 Jun 2017 17:07:25 GMT Rollbar:log Successful api response. Link: https://rollbar.com/occurrence/uuid/?uuid=25e5e414-8588-4024-d2c6-de3c810f80b6
=> Exited with code: 1
=> Your application is crashing. Waiting for file change.
```
Here is what I configure:
```
export DEBUG=Rollbar:* node 
```

So I see the log message from Rollbar, but I do not see the actual exception that caused it.
Any idea how to get the actual exception stack trace in the log?

tophernuts:
https://github.com/rollbar/rollbar.js/issues/309

The verbose option was just recently added back, but there hasn't been a release yet, I think that's why you're not seeing the stack trace yet!

gerwinbrunner:
When is this going to be released?

Mit freundlichen Grüßen
Gerwin Brunner

rokob:
v2.1.0 should resolve this

PR #332, opened 2017-07-06
Fix #331 Verbose logging was using the wrong key structure

### LINKED ISSUE #331: Verbose logging does not work
When passed the `verbose: true` option, rollbar.js does not log errors to the console.

In `queue.js:221`, the message to log is built using the following:

```
message = message || _.get(item, 'data.body.trace_chain.0.exception.message');
```

Digging in to this, `item` does not have a `data` property (`item.body` does exist, however), so this ends up being undefined and nothing is logged. The passing tests do not seem to reflect the actual objects being passed to this function.

Once this is corrected, all that is logged to the console is the error _message_, with no trace information. With `node-rollbar`, we got logs like this:

```
Rollbar:error Handling uncaught exception. +0ms
  Rollbar:error TypeError: Cannot read property 'work' of null
    at Object.<anonymous> (/Users/andy/XXXX/src/XXXX.ts:58:2)
    at Module._compile (module.js:541:32)
    at Module.m._compile (/Users/andy/XXXX/node_modules/ts-node/src/index.ts:413:23)
    at Module._extensions..js (module.js:550:10)
    at Object.require.extensions.(anonymous function) [as .ts] (/Users/andy/XXXX/node_modules/ts-node/src/index.ts:416:12)
    at Module.load (module.js:456:32)
    at tryModuleLoad (module.js:415:12)
    at Function.Module._load (module.js:407:3)
    at Function.Module.runMain (module.js:575:10)
    at Object.<anonymous> (/Users/andy/XXXX/node_modules/ts-node/src/_bin.ts:182:12)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:456:32)
    at tryModuleLoad (module.js:415:12)
    at Function.Module._load (module.js:407:3)
    at Function.Module.runMain (module.js:575:10) +1ms
```

which we would like to preserve after upgrading.

rokob:
The queue gets the item to log that represents the data in the final payload structure of
```
{
  "access_token": "abc123",
  "data": { ... }
}
```

rather than the whole payload itself. The API object is responsible for wrapping that data object up
inside a structure like the above and then sending it. Thus the body of the item is a top level key
rather than nested under the data key.

Secondly, everyone wants the original error to be output if there is one isntead of just the
exception message, so make it so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -994,6 +994,18 @@ If you're using the [Passport](http://passportjs.org/) authentication library, t
 
 Note: in Rollbar, the `id` is used to uniquely identify a person; `email` and `username` are supplemental and will be overwritten whenever a new value is received for an existing `id`. The `id` is a string up to 40 characters long.
 
+## Verbose Option
+
+If you would like to see what is being sent to Rollbar in your console, use the
+`verbose` option. Set `verbose: true` in your configuration, and we will output certain information
+via the [debug](https://www.npmjs.com/package/debug) package. This package uses the `DEBUG`
+environment variable to configure what to output. We use the namespace `Rollbar` for our
+log messages, so for example, to see everything you need to do something like this:
+
+```
+DEBUG=Rollbar:* node app.js
+```
+
 ## Upgrading from node_rollbar
 
 The upgrade path from `node_rollbar` version 0.6.4 to version 2.0.0 of this library is not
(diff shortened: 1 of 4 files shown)


rokob:
This also fixes https://github.com/rollbar/rollbar.js/issues/320

Issue #302, opened 2017-06-02
Provide TypeScript declaration files


dlindenkreuz:
See https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html

coryvirok:
Thanks @dlindenkreuz. We don't currently have plans to do this but are certainly open to PRs!

TimMensch:
FYI, lack of TypeScript declarations is making investigate other options. Been using and liking Rollbar, but I'm not ready to volunteer to create and maintain TypeScript declarations for a commercial project.

Also: You can hardly claim Angular 2 support with declarations, given that Angular 2 recommends TypeScript.

emk:
Having TypeScript declarations would be a huge win for my employer; rollbar is the last server-side library that we use regularly that we still don't have type declarations for. Thank you to @rokob for working on this!

PR #334, opened 2017-07-11
first pass at a typescript declaration file

rokob:
Addresses #302

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,48 @@
+// Type definitions for rollbar 2.1.1
+// Project: Rollbar
+
+/*~ This is the module template file for class modules.
+ *~ You should rename it to index.d.ts and place it in a folder with the same name as the module.
+ *~ For example, if you were writing a file for "super-greeter", this
+ *~ file should be 'super-greeter/index.d.ts'
+ */
+
+export = Rollbar;
+
+declare class Rollbar {
+    constructor(options: Rollbar.Configuration?);
+    static init(options: Rollbar.Configuration): Rollbar;
+
+    public global(options: Rollbar.Configuration): Rollbar;
+    public configure(options: Rollbar.Configuration): Rollbar;
+    public lastError(): Error?;
+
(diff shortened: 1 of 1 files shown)


brianr:
Nice!

fyi @TimMensch that this is in progress

TimMensch:
Thanks for the update and quick response. :)

rokob:
I'm going to merge this so there is something and we can iterate on it further if there are any issues.

index.d.ts
```diff
@@ -0,0 +1,48 @@
...
+/*~ This is the module template file for class modules.
+ *~ You should rename it to index.d.ts and place it in a folder with the same name as the module.
+ *~ For example, if you were writing a file for "super-greeter", this
+ *~ file should be 'super-greeter/index.d.ts'
+ */
+
+export = Rollbar;
+
+declare class Rollbar {
+    constructor(options: Rollbar.Configuration?);
```
Agamnentzar: should be `constructor(options: Rollbar.Configuration | undefined);`

index.d.ts
```diff
@@ -0,0 +1,48 @@
...
+    static init(options: Rollbar.Configuration): Rollbar;
+
+    public global(options: Rollbar.Configuration): Rollbar;
+    public configure(options: Rollbar.Configuration): Rollbar;
+    public lastError(): Error?;
```
Agamnentzar: should be `public lastError(): Error | undefined;`

index.d.ts
```diff
@@ -0,0 +1,48 @@
...
+        logLevel?: Level;
+        reportLevel?: Level;
+        uncaughtErrorLevel?: Level;
+        endpoint?: string;
+        verbose?: boolean;
+        enabled?: boolean;
+        captureUncaught?: boolean;
+        captureUnhandledRejections?: boolean;
+    }
+    export type Callback = (err: Error?, response: object) => void;
```
Agamnentzar: should be `export type Callback = (err: Error | undefined, response: object) => void;`

PR #335, opened 2017-07-11
[Fixes #269] get rid of rollbar wrapper stack frames

### LINKED ISSUE #269: Rollbar event wrapper should have nicer stack traces
We wrap certain functions such as certain global event handlers in order to catch otherwise unhandled exceptions that do not necessarily go through the window.onerror handler. This leads to stack traces with what looks like rollbar code causing the error due to the _wrapped method showing up at the top of the stack trace.

Additionally, this makes host whitelisting hard to do for bundled code because that _wrapped frame will usually have a filename that matches the client's bundle because that is where rollbar lives.

The code that does the wrapping should pass information to rollbar.error which signals that the error is being generated within a wrapped function. This can then be used to modify that particular frame in the trace to have something like:

```js
{
    lineno: null,
    colno: null,
    method: null,
    filename: "<Rollbar Event Wrapper>"
}
```

rokob:
We wrap certain built-ins in the browser and rethrow exceptions. When such an error occurs, our wrapper ends up in the stack trace because technically it is running, but it isn't really part of the error so there is no reason for it to be in the stack trace.

There are a few different ways I looked at fixing this, but in the end just checking for our `f._wrapped` method signature in the frame is the simplest and actually works about as well as any other approach. Threading state around is a bit of a mess, and actually munging the exception at the rethrow site is impossible due to browser differences.

### DIFF
--- a/src/browser/transforms.js
+++ b/src/browser/transforms.js
@@ -172,7 +172,7 @@ function addBodyTrace(item, options, callback) {
         method: (!stackFrame.func || stackFrame.func === '?') ? '[anonymous]' : stackFrame.func,
         colno: stackFrame.column
       };
-      if (frame.method && frame.method.indexOf('f._wrapped') != -1) {
+      if (frame.method && frame.method.endsWith && frame.method.endsWith('._rollbar_wrapped')) {
         continue;
       }
 
--- a/src/browser/transforms.js
+++ b/src/browser/transforms.js
@@ -172,6 +172,9 @@ function addBodyTrace(item, options, callback) {
         method: (!stackFrame.func || stackFrame.func === '?') ? '[anonymous]' : stackFrame.func,
         colno: stackFrame.column
       };
+      if (frame.method && frame.method.indexOf('f._wrapped') != -1) {
+        continue;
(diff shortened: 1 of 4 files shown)


src/browser/transforms.js
```diff
@@ -172,6 +172,9 @@ function addBodyTrace(item, options, callback) {
         method: (!stackFrame.func || stackFrame.func === '?') ? '[anonymous]' : stackFrame.func,
         colno: stackFrame.column
       };
+      if (frame.method && frame.method.indexOf('f._wrapped') != -1) {
```
brianr: This seems fragile - it might break if:

- this library changes, so the minified frame is no longer called `f._wrapped`
- this library is included in a larger application build, and the minified name there is something else

How about if we change the `_wrapped` part to be more identifiable? e.g. `_rollbar_wrapped` -- and then we can just check that the method name ends with the exact string `._rollbar_wrapped`

brianr:
👍 

Will this require people to update their snippet?

rokob:
Yeah, the snippet will change

PR #337, opened 2017-07-12
[Fixes #336] Propagate message

### LINKED ISSUE #336: Confused About logger.error API
Per the documentation, you can pass into `rollbar.error` an exception, a string, and an object. What I'd like to do is pass an exception and a string that's a simple human readable label. However, it seems like if I pass an exception into `rollbar.error`, the string is ignored and only the exception information is logged. I can pass an object in with meta data and get around this, but that seems counter to the documentation.

I'm currently trying with `logger.error(result, 'error retrieving databases');` where `results` is an instance of `Error`.

rokob:
Include message if we have one even if we also have a stack trace


rokob:
Github is screwed up, re: #336 the title of this PR should link these two but it doesn't

PR #338, opened 2017-07-14
add some intervals to keep the event loop running when wait is called

rokob:
It is possible that calling wait does not actually keep the event loop spinning while all events are processing. The proper use of interval timers in the queue will keep things open until all the pending requests have actually gone out.


rokob:
Merging

coryvirok:
Whoa. This is bad-ass.

PR #340, opened 2017-07-19
Add Telemetry Data

rokob:
The purpose of this PR is to enhance events sent to Rollbar with telemetry data. This is a work in progress.


src/telemetry.js
```diff
@@ -0,0 +1,100 @@
...
+    level: getLevel(type, level),
+    type: type,
+    timestamp_ms: now(),
+    metadata: metadata
+  };
+  this.push(e);
+  return e;
+};
+
+Telemeter.prototype.captureError = function(err) {
```
coryvirok: It's a bit weird to have this as well as `captureRollbar` but I can see a reason to have a distinction. What if we added an optional Rollbar param to each of these methods to link them with a Rollbar occurrence if one was sent for the event?

E.g.

```js
Telemeter.prototype.captureError = function(err, rollbar_uuid=null) {
  //...
}
Telemeter.prototype.captureLog = function(err, rollbar_uuid=null) {
  //...
}

//...
```

This way we allow the addition of telemetry items which are and are not associated with Rollbar occurrences.

src/browser/telemetry.js
```diff
@@ -1,15 +1,15 @@
 var _ = require('../utility');
 
 var defaults = {
-	network: true,
-	console: true,
-	dom: true,
-	location: true
+  network: true,
+  console: true,
```
coryvirok: Can we rename this to "log" to be more backend-friendly?

src/browser/telemetry.js
```diff
@@ -1,15 +1,15 @@
...
+  dom: true,
+  location: true
```
coryvirok: Oh, I can imagine we may want to allow users to capture gps location data but this key is for the nav location, correct? Maybe we should rename that to "navigation"?

src/browser/telemetry.js
rokob: 👍

src/browser/telemetry.js
rokob: Yeah navigation makes more sense. The name comes from `window.location`

src/browser/shim.js
```diff
@@ -112,7 +112,7 @@ Shim.prototype.loadFull = function(window, document, immediate, options, callbac
   parentNode.insertBefore(s, f);
 };
 
-Shim.prototype.wrap = function(f, context) {
+Shim.prototype.wrap = function(f, context, _before) {
```
coryvirok: Looks like Shim.prototype.wrap and Rollbar.prototype.wrap are almost identical. Can we consolidate?

src/browser/telemetry.js
```diff
@@ -0,0 +1,228 @@
...
+}
+
+function Instrumenter(options, telemeter, rollbar, _window, _document) {
+  var autoInstrument = options.autoInstrument;
+  if (autoInstrument === false) {
+    this.autoInstrument = {};
+    return;
+  }
+  if (!_.isType(autoInstrument, 'object')) {
+    autoInstrument = defaults;
```
coryvirok: Can we log a warning or something here? It could help debugging configuration issues.

rokob: The idea here is that the configuration option `autoInstrument` can be either a boolean `true | false` or an object with possible keys `network`, `log`, `dom`, `navigation`, `connectivity`. If the option is set to `false` then we do not instrument anything. If the value is set to `true` then we instrument everything based on the default values. Otherwise, we merge the object that is passed in with the defaults.

So:

```
{
    autoInstrument: false
}
=> Instrument nothing
```

```
{
    autoInstrument: true
}
<=>
{
    autoInstrument: {
        network: true,
        log: true,
        dom: true,
        navigation: true,
        connectivity: true
    }
}
```

```
{
    autoInstrument: {
        log: false,
        navigation: false
    }
}
=>
{
    autoInstrument: {
        network: true,
        log: false,
        dom: true,
        navigation: false,
        connectivity: true
    }
}
```

coryvirok: Ah, that makes sense and looks pretty useful.

src/telemetry.js
```diff
@@ -0,0 +1,94 @@
+var _ = require('./utility');
+
+function Telemeter(options) {
+  this.queue = [];
+  this.options = _.extend(true, {}, options);
+  var maxTelemetryEvents = this.options.maxTelemetryEvents;
+  // pin the queue size to the interval [0, 100]
+  this.maxQueueSize = Math.max(0, Math.min(maxTelemetryEvents || 100, 100))
```
coryvirok: Let's move the `100` to a module constant

src/telemetry.js
```diff
@@ -0,0 +1,94 @@
...
+  return e;
+};
+
+Telemeter.prototype.captureEvent = function(metadata, level) {
+  return this.capture('manual', metadata, level);
+}
+
+Telemeter.prototype.captureError = function(err) {
+  return this.capture('error', {
+    message: err.message || String(err)
```
coryvirok: We should also include the stack trace string from `err.stack` if it exists.

src/telemetry.js
```diff
@@ -0,0 +1,94 @@
...
+Telemeter.prototype.captureLog = function(message, level) {
+  return this.capture('log', {
+    message: message
+  }, level);
+};
+
+Telemeter.prototype.captureNetwork = function(metadata, subtype) {
+  subtype = subtype || 'xhr';
+  metadata.subtype = metadata.subtype || subtype;
+  return this.capture('network', metadata, 'info');
```
coryvirok: If this is an XHR event, can we get the status code and change the level based on it?

src/telemetry.js
```diff
@@ -0,0 +1,94 @@
...
+};
+
+Telemeter.prototype.captureRollbar = function(item) {
```
coryvirok: I think we should remove this and have an optional Rollbar UUID for each of the other capture methods. The idea being that we would be able to associate a Rollbar occurrence with any of the telemetry objects.

src/telemetry.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+  return this.capture('dom', metadata, 'info');
+};
+
+Telemeter.prototype.captureNavigation = function(from, to) {
+  return this.capture('navigation', {from: from, to: to}, 'info');
+};
+
+Telemeter.prototype.captureConnectivityChange = function(type) {
+  return this.capture('connectivity', {change: type}, 'info');
```
coryvirok: Should this be a `network` event with a subtype `connectivity`?

src/telemetry.js
```diff
@@ -0,0 +1,94 @@
...
+    this.queue.shift();
+  }
+};
+
+function getLevel(type, level) {
+  if (level) {
+    return level;
+  }
+  var defaultLevel = {
+    'error': 'error',
```
coryvirok: Style: no need for quotes around the key names.

src/utility.js
```diff
@@ -534,6 +534,30 @@ function _getScrubQueryParamRegexs(scrubFields) {
   return ret;
 }
 
+function formatArgsAsString(args) {
+  var i, len, arg;
+  for (i=0, len=args.length; i < len; i++) {
```
coryvirok: Style: Add spaces around operators.

PR #341, opened 2017-07-21
unify configuration options

rokob:
The configuration options for capturing unhandled exceptions and rejections should be the same on the browser and server


rokob:
This allows both of the old options on both the browser and server to ensure backwards compatibility, but updates the README to the desired configuration option. We can then remove those extra options later.

coryvirok:
Yes! Thanks for this.

PR #345, opened 2017-07-27
Address issue with handleItemWithError possible not calling the callback

rokob:
fix one case where we might have called the callback twice which could cause odd behaviour


rokob:
I traced all of the code paths through the transforms and parsing errors versus messages. The only thing I can see that could have been causing issues was a certain callback getting called twice which could have caused an issue with the way the indexing is shared in the transforms. Exactly what this would do is not clear, but it is nonetheless wrong and this fixes it.

PR #346, opened 2017-07-28
Allow more flexibility of request objects

rokob:
Allow more flexibility in the type of request objects we can handle on the server side. Specifically, handle the type of requests that hapi.js creates


epikhighs:
@rokob when will this fix be tagged for a release?

PR #356, opened 2017-08-03
document nojson dist for CSP

rivkahstandig3636:


brianr:
Going to revert this pending discussion. @rivkahstandig3636 can you test whether anything special is needed regarding CSPs?

PR #365, opened 2017-08-08
Add accessToken to typescript configuration interface

martynsmith:
This seems like it was accidentally left out of the typescript definition?


rokob:
:+1:

PR #375, opened 2017-08-14
update the level when we find out the status

rokob:


rokob:
We "capture" the event when the request starts, and then we use a private reference on the xhr object to hold a reference into the telemetry event as the request progresses. This allows us to update certain data while the request is inflight. The problem was that we were setting the level based on the status code at the beginning, updating the status code when the request completes, but not updating the subsequent level.

PR #377, opened 2017-08-16
Use JSON polyfill if the detected JSON.* arent native funtions

kjones1876:
The prototype framework (and probably more) apply JSON polyfills which arent standards compliant leading to API rejections. If the JSON.* arnt native, use our own polyfill.

### DIFF
--- a/src/utility.js
+++ b/src/utility.js
@@ -9,10 +9,10 @@ function setupJSON() {
   __initRollbarJSON = true;
 
   if (isDefined(JSON)) {
-    if (isFunction(JSON.stringify)) {
+    if (isNativeFunction(JSON.stringify)) {
       RollbarJSON.stringify = JSON.stringify;
     }
-    if (isFunction(JSON.parse)) {
+    if (isNativeFunction(JSON.parse)) {
       RollbarJSON.parse = JSON.parse;
     }
   }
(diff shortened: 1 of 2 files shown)


src/utility.js
```diff
@@ -71,6 +71,19 @@ function isFunction(f) {
...
+ * @param f - any value
+ * @returns true if f is a native JS function, otherwise false
+ */
+function isNativeFunction(f) {
+  if (!isFunction(f)) {
+    return false;
+  }
+  var functionString = Function.prototype.toString.call(f);
+  return /\{\s+\[native code\]/.test(functionString);
+}
```
rokob: What browsers does this work and not work in?

kjones1876: I tested it on IE9 and it works, however im struggling to find any coverage reports on the usage of `[native code]`. I've been trying to test it in IE6 manually but i cant open firebug using browser stack. Do you have an cross browse tests i can write ?

kjones1876: @rokob ^

chrisbarmonde: FWIW, `lodash` has an `isNative` function that might be more robust. It essentially does a check for `[native code]`, but it does it by actually getting the toString result of a native code function and doing a bunch of regex magic to turn it into a regex to match against other native functions, probably to make up for any minor differences between browsers.

https://github.com/lodash/lodash/blob/master/isNative.js

Issue #379, opened 2017-08-16
Network telemetry blocks Pusher from subscribing to channels


EtienneLem:
Haven’t been able to find exactly _where_ or _why_ it stops Pusher from working, but setting `autoInstrument: { network: false }` fixes the issue for me.

From what I can tell, the Pusher client still successfully sends XHR requests to my server to get an authentication token for the channels I want to subscribe to (see [Authenticating users](https://pusher.com/docs/authenticating_users)), but the `subscription_succeeded` never gets fired anymore.

Disabling network telemetry isn’t a problem for me, but I thought I’d bring it to your attention nonetheless.

emkman:
After upgrading to 2.2, Rollbar broke the callback of the Facebook sdk for us. This broke both login and new user signup for us and fails silently (no error, no console output, nothing logged as a Rollbar item. Shame on me for trusting this would be production ready, but it looks like there are still some serious issues with Telemetry and I question it being enabled by default. Reverting back to 2.1.3 or 2.0.4 both resolved the issue immediately.

jessewgibbs:
@EtienneLem @emkman thanks for bringing this to our attention, and sorry for the inconvenience.  We've begun investigating this and will comment on this issue to keep you up-to-date.

jessewgibbs:
We've just released [rollbar.js 2.2.1](https://github.com/rollbar/rollbar.js/releases/tag/v2.2.1) to disable the default network instrumentation while we investigate this issue.

fraserredmond:
Something similar for us, an XMLHttpRequest.onreadystatechange event wasn't firing - so maybe it's getting overwritten by your event handler.

jessewgibbs:
@fraserredmond thanks for the info.  As Etienne mentioned above, you can set `autoInstrument: { network: false } ` to disable the network event capture until we get this sorted out.

brianr:
This should be fixed now, as of today's 2.2.3 release. Going to close this issue; see #382 for details of the fix.

@EtienneLem, @emkman, and @fraserredmond - mind trying out the latest version and see if the issue is fixed?

PR #382, opened 2017-08-17
fix telemetry network bug

rokob:
The third argument to replace is a function that is required to return a function which becomes the new function for a given property on an object. If you don't return a function then the property on the object gets replaced with undefined, breaking everything.

This only manifests if onreadystatechange is defined on an xhr request, which was not something I did during manual or automated testing. I added it to the manual test page and verified that before this diff things were broken, and afterwards everything works as expected.


rokob:
Fixes #379

PR #387, opened 2017-08-18
explain more about how telemetry works

rokob:


README.md
```diff
@@ -363,6 +363,25 @@ _rollbarConfig = {
 }
 ```
 
+The different types of events that we automatically capture are: `network`, `log`, `dom`,
+`navigation`, and `connectivity`.
+
+Network events are XHR and fetch requests. We store the status code, the url, and some timing events
+to determine how long requests take.
+
+Log events are calls to `console` and we simple store which console method was called and the
```
rokob: s/we simple store/we simply store/

README.md
```diff
@@ -372,6 +391,43 @@ var level = 'info'; // Possible values: 'debug', 'info', 'warning', 'error', 'cr
...
+  type: string
+  timestamp_ms: number
+  body: object
+  source: string
+  uuid?: string
+}
+```
+
+The size of each of these events is mostly determined by the `body` field, however we attempt to
+store only the smallest amount of information necessary aide in understanding. Therefore, if you
```
rokob: s/aide/to aid/

PR #389, opened 2017-08-19
no matter what Date.now() returns, make it a number

### LINKED ISSUE #388: API error: Invalid format. data.body.telemetry[0].timestamp_ms string value found, but a integer is required.
I upgraded our rollbar.js to 2.2.3 today (using the snippet defined on the README). Since then we've been getting errors from the "Rollbar System" that say

API error: Invalid format. data.body.telemetry[0].timestamp_ms string value found, but a integer is required. data.body.telemetry[1].timestamp_ms string value found, but a integer is required. data.body.telemetry[2].timestamp_ms string value found, but a integer is required. data.body.telemetry[3].timestamp_ms string value found, but a integer is required.

rokob:
Fixes #388 

If `Date.now()` returns `new Date()` which is what Date.js does for some crazy reason, then `+Date.now()` is the same as `+new Date()` which is the same fallback we have when `Date.now` is undefined. Otherwise, if `Date.now()` returns a timestamp like it should, then `+Date.now()` is just the same as `Date.now()` through the magic of Javascript operators.

### DIFF
--- a/src/utility.js
+++ b/src/utility.js
@@ -557,7 +557,7 @@ function formatArgsAsString(args) {
 
 function now() {
   if (Date.now) {
-    return Date.now();
+    return +Date.now();
   }
   return +new Date();
 }


src/utility.js
```diff
@@ -557,7 +557,7 @@ function formatArgsAsString(args) {
 
 function now() {
   if (Date.now) {
-    return Date.now();
+    return +Date.now();
```
brianr: Maybe add a comment here with a link to this issue?

PR #392, opened 2017-08-21
Ensure hostBlackList is considered when filtering errors

chrisrbnelson:
The ability to black list hostnames was [added recently](https://github.com/rollbar/rollbar.js/pull/319) but doesn't appear to actually be used.

I'm not overly familiar with the Rollbar source, but from what I can see it simply needs to be added to the list of predicates (in the inverse form, so it returned true if not blacklisted).

### DIFF
--- a/src/browser/predicates.js
+++ b/src/browser/predicates.js
@@ -141,7 +141,6 @@ function messageIsIgnored(item, settings) {
 module.exports = {
   checkIgnore: checkIgnore,
   userCheckIgnore: userCheckIgnore,
-  urlIsBlacklisted: urlIsBlacklisted,
   urlIsNotBlacklisted: urlIsNotBlacklisted,
   urlIsWhitelisted: urlIsWhitelisted,
   messageIsIgnored: messageIsIgnored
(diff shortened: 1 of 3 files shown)


src/browser/rollbar.js
```diff
@@ -360,6 +360,7 @@ function addPredicatesToQueue(queue) {
   queue
     .addPredicate(predicates.checkIgnore)
     .addPredicate(predicates.userCheckIgnore)
+    .addPredicate(predicates.urlIsNotBlacklisted)
```
rokob: 👍

src/browser/predicates.js
```diff
@@ -36,6 +36,10 @@ function urlIsBlacklisted(item, settings) {
   return urlIsOnAList(item, settings, 'blacklist');
 }
 
+function urlIsNotBlacklisted(item, settings) {
+  return !urlIsBlacklisted(item, settings);
```
rokob: Rather than have this indirection, just replace `!urlIsBlacklisted(...)` with `!urlIsOnAList(item, settings, 'blacklist')` and get rid of the `urlIsBlacklisted` function.

chrisrbnelson: @rokob Good call - pushed up that change.

PR #403, opened 2017-08-23
Use lodash is native function checking for JSON.* functions

kjones1876:
@rokob replaces #377


rokob:
This seems reasonable to me

PR #405, opened 2017-08-25
Replace `object` with `Object` in TypeScript definition

artnc:
Before making this change, I was getting TS errors like:
```
Error in /home/art/git/duolingo-web/node_modules/rollbar/index.d.ts
(54,48): error TS2304: Cannot find name 'object'.
```
This appears to be a [known issue](https://github.com/Microsoft/TypeScript/issues/14647) affecting newer TS versions.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -25,7 +25,7 @@ declare class Rollbar {
     public error(...args: Rollbar.LogArgument[]): Rollbar.LogResult;
     public critical(...args: Rollbar.LogArgument[]): Rollbar.LogResult;
 
-    public captureEvent(metadata: object, level: Rollbar.Level): Rollbar.TelemetryEvent;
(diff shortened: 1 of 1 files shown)


rokob:
This will then still work with older TS versions right?

artnc:
My mistake - it's actually all versions older than 2.2 that are broken right now since `object` was [introduced](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#object-type) six months ago in 2.2.

If you want to support TS versions <2.2 then `object` will need to be removed, but I think it's also fair to just close this PR since the TS team [suggests](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html) `object` going forward.

rokob:
I'm going to close this because of the official suggestion to use `object` going forward.

PR #414, opened 2017-08-28
we should not be storing the props for every request, only the prototype one

### LINKED ISSUE #413: Updating to 2.2.3 with Telemetry enabled is gobbling up browser memory
We recently updated our rollbar backend SDK and front end to 2.2.3 to take advantage of telemetry data.   

Since updating, we noticed that browser memory used by our app was accumulating each time a user changed a view within our product (our views are React based interactive - lane kanban type views with saved filters). 

When we disabled Telemetry, the memory consumption issues go away.  I've prepared a heap dump from Google chrome to show the differences we are seeing in a short moment between switching views and what is going on with the rollbar instrumentor.  

You can get it here along with the screenshots:

https://www.dropbox.com/s/r1v4iummg5doazc/heap.zip?dl=0

rokob:
Fixes #413 

There is something like a memory leak, at least it has the same effect if it is not classically a leak. We were storing wrapped properties for every network request rather than just the wrapped properties on the XHR prototype. While this is maybe technically correct in some abstract sense as it allows us to properly unwrap everything if the configuration changes, we don't really need to unwrap those for things to still work. Also storing them will result in a growing memory footprint without any real cleanup mechanism. The alternative to this fix is to have the onreadystatechange handler do some cleanup when the request is done, but this is fix is adequate.

### DIFF
--- a/src/browser/telemetry.js
+++ b/src/browser/telemetry.js
@@ -113,7 +113,7 @@ Instrumenter.prototype.instrumentNetwork = function() {
     if (prop in xhr && _.isFunction(xhr[prop])) {
       replace(xhr, prop, function(orig) {
         return self.rollbar.wrap(orig);
-      }, self.replacements, 'network');
+      });
     }
   }
 


alanosman:
nice - glad you guys caught this problem.

alanosman:
Any ideas what version it will make it's way into?

PR #419, opened 2017-09-01
Use json polyfil if non native

kjones1876:
A version of #403 without using lodash so theres no shim protection.

### DIFF
--- a/src/utility.js
+++ b/src/utility.js
@@ -9,10 +9,10 @@ function setupJSON() {
   __initRollbarJSON = true;
 
   if (isDefined(JSON)) {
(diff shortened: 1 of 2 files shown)


rokob:
This is much better than bringing in lodash for just this one function.

kjones1876:
Originally the way I did would have only included the relevant parts of lodash, not the whole library. Just I didn't realise that there's a statement checking for old versions of core-js. The libraries compiler seems to do a lot. If you check the isNative.js file in the repo it's not in there but that's not the isNative.js file is the npm package

PR #422, opened 2017-09-05
if enabled is false then set all instrumentation to false

### LINKED ISSUE #400: Enabling Telemetry functionality changes the script/line number reported in browser console.
Normally when using window.console methods, the message logged to the console window contains the script file and line number that generated the log message.  When you enable the Telemetry functionality in Rollbar, any log messages are now reported as coming from the Rollbar script (and line number) and not the originating script/line number.

e.g.
rollbar-telemetry

rokob:
Fixes #400

### DIFF
--- a/src/browser/shim.js
+++ b/src/browser/shim.js
@@ -55,10 +55,12 @@ function setupShim(window, options) {
     }
 
     var ai = options.autoInstrument;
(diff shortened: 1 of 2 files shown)


gaiottino:
When will this make it's way into a release? What release should I look out for?

gaiottino:
Any update which release (and when) this will be available?

PR #431, opened 2017-09-18
Fix missing argument name in TypeScript definition file

timstott:
Fixes the invalid type definition of the newly added `lambdaHandler`


rokob:
👍

PR #435, opened 2017-09-19
fix the umd build to not pollute the global namespace

### LINKED ISSUE #433: UMD build sets window.global = () => {...}, messing up my app
in the rollbar UMD build, i see this: `r.global=function(t){return w?w.global(t):void o()}`. what's up with that? if i load it in the browser, i see:

screen shot 2017-09-18 at 5 18 47 pm

should i be using a different file for a global `Rollbar`?

rokob:
Fixes #433

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -112,6 +112,7 @@ var UMDConfigBase = {
   },
   output: {
     path: outputPath,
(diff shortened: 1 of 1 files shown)


jonathanong:
👍

PR #445, opened 2017-09-26
Require JSON3 dependency with absolute path

timstott:
Cannot bundle Rollbar with Webpack 3 since it cannot require `json3.js` ✨ .

```
ERROR in ./node_modules/rollbar/src/utility.js
Module not found: Error: Can't resolve '../vendor/JSON-js/json3.js' in '__ABS_PATH__/node_modules/rollbar/src'
 @ ./node_modules/rollbar/src/utility.js 21:26-63
 @ ./node_modules/rollbar/src/server/rollbar.js
 @ ./src/rollbar.ts
 @ ./src/handlers.ts
```

giphy-downsized 1

### DIFF
--- a/src/utility.js
+++ b/src/utility.js
@@ -1,4 +1,5 @@
 var extend = require('extend');
+var path = require('path');
 
 var RollbarJSON = {};
 var __initRollbarJSON = false;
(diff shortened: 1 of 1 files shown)


rokob:
There is no path module in the browser

timstott:
In the end I added an alias to the resolve webpack config. Good enough solution for the time being 👌 

```
module.exports = {
  resolve: {
    alias: {
      rollbar: 'rollbar/dist/rollbar.min.js'
    },
   ...
}

```

PR #455, opened 2017-10-06
Update README.md

rubenCodeforges:
For Angular2+ projects with aot on, the example would run users in to an error , arrow functions are not supported in aot.

The next issue would be the ErrorHandler , in the example it would override the existing ErrorHandler which would remove console.error , adding it. The naming of the ErrorHandler changed, to reflect its purpose.

Next we create a wrapper service for Rollbar to exclude naming conflicts and avoid usage of the injector.
Also added some information about npm installation of the lib.


rokob:
Looks great to me, I have tested this with Angular 4, it doesn't appear to use anything that was not in Angular 2 so I am pretty sure it works there as well.

rubenCodeforges:
@rokob Ive took it out of a Angular4 project, so it was tested.

rokob:
The README has been updated to include basically these same changes. I'm leaning towards closing this as I think after rebasing there isn't really a change left.

rubenCodeforges:
@rokob ive rebased and reviewed the changes in the README in general its all the same , as i stated before , injectors are not required , you can utilize a service for this , this is much more cleaner approach.

The troubleshoot section is not required as well as my changes will not cause compilation errors with `--aot` and AOT is always used for staging and production builds.

rokob:
Can we rebase this so it is just the README change?

rubenCodeforges:
@rokob obviously this PR is not useful for you guys, im closing it

PR #466, opened 2017-10-17
Fix global config

rokob:
Working out the kinks in the way maxItems is handled throughout the lifecycle of the notifier.

### DIFF
--- a/src/rollbar.js
+++ b/src/rollbar.js
@@ -43,6 +43,7 @@ Rollbar.prototype.configure = function(options, payloadData) {
     payload = {payload: payloadData};
   }
   this.options = _.extend(true, {}, oldOptions, options, payload);
+  this.global(this.options);
   return this;
 };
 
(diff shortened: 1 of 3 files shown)


src/rollbar.js
```diff
@@ -14,6 +14,7 @@ var _ = require('./utility');
 function Rollbar(options, api, logger, platform) {
   this.options = _.extend(true, {}, options);
   this.logger = logger;
+  Rollbar.rateLimiter.configureGlobal(this.options);
```
rokob: This fixes the issue where the maxItems limit was not getting set via the constructor and therefore from the global _rollbarConfig.

src/rollbar.js
```diff
@@ -42,6 +43,7 @@ Rollbar.prototype.configure = function(options, payloadData) {
     payload = {payload: payloadData};
   }
   this.options = _.extend(true, {}, oldOptions, options, payload);
+  this.global(this.options);
```
rokob: This allows global settings to come through configure which I suspect people are also doing.

PR #468, opened 2017-10-18
Fix: calling Rollbar.configure shouldn't clear telemetry config.

brianmcd:
Let me know if this is just a case of me doing it wrong, but here are the repro steps to what I was seeing:

1.) Initially configure Rollbar with `maxTelemetryEvents: 15` or something that isn't `100`.
2.) At some point in the future, call `Rollbar.configure`, e.g.
```javascript
Rollbar.configure({
  payload: {
    person: { id: 1, email: 'me@example.com' }
  }
});
```
__Expected__: `maxTelemetryEvents` is still set to 15.
__Actual__: `maxTelemetryEvents` is reset to the default (100).

### DIFF
--- a/src/telemetry.js
+++ b/src/telemetry.js
@@ -10,7 +10,8 @@ function Telemeter(options) {
 }
 
 Telemeter.prototype.configure = function(options) {
-  this.options = _.extend(true, {}, options);
+  var oldOptions = this.options;
+  this.options = _.extend(true, {}, oldOptions, options);
   var maxTelemetryEvents = this.options.maxTelemetryEvents || MAX_EVENTS;
   var newMaxEvents = Math.max(0, Math.min(maxTelemetryEvents, MAX_EVENTS));
   var deleteCount = 0;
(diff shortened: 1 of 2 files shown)


src/telemetry.js
```diff
@@ -10,7 +10,8 @@ function Telemeter(options) {
 }
 
 Telemeter.prototype.configure = function(options) {
-  this.options = _.extend(true, {}, options);
+  var oldOptions = this.options;
+  this.options = _.extend(true, {}, oldOptions, options);
```
brianmcd: I just copied how the notifier does it: https://github.com/rollbar/rollbar.js/blob/b6a3fce240bee0be185cf8e9982f321589bba8dc/src/notifier.js#L26-L27

test/telemetry.test.js
```diff
@@ -65,4 +65,21 @@ describe('configure', function() {
     expect(t.queue.length).to.equal(5);
     done();
   });
+  it('does not drop existing options that are not passed to configure', function(done) {
```
rokob: Nice 💯

brianmcd:
Thanks for the quick merge!

brianmcd:
@rokob - Do you know when this change will make it into a release?

PR #472, opened 2017-10-23
Fix log when verbose is active

guiconti:
I manage to fix changing the console bind to logger on logger.js


rokob:
Closing in favor of #473

Issue #470, opened 2017-10-23
verbose: true not logging to console


guiconti:
Hi, we are using rollbar on one of our node.js projects and we can not make it log info to console when we call debug, log, info, warning, error and critical. Here is where we declare rollbar:

```
const Rollbar = require('rollbar');
const rollbar = new Rollbar({
  accessToken: process.env.ROLLBAR,
  verbose: true,
  captureUncaught: true,
  captureUnhandledRejections: true,
  payload: {
    environment: process.env.NODE_ENV
  }
});
rollbar.error('Why i don`t go to console?');
```

Is something wrong with the verbose option? We are using version 2.2.10 with node.js v8.4.0

Thank you!

guiconti:
I managed to fix it. I opened a PR #472

rokob:
https://github.com/Rollbar/rollbar.js/#verbose-option-1

rokob:
For node.js, verbose needs to be true AND you need to configure the debug package via an environment variable to log the output. This is how node_rollbar always worked and how rollbar.js on the server side has always worked. This has come up enough that I am now convinced this is overly confusing. I'm open to not using the debug module and just having the behaviour be that `verbose: true` implies logging to the console. I am not sure the reason behind using the debug module in the first place. Anyone out there opposed to that?

guiconti:
@rokob I submitted a PR #473 that remove the need to use debug for verbose loggin on node.js. Hope it helps.

dozoisch:
Getting the same issue on the web. Seems like `enabled: false` with `verbose: true` doesn't log anymore.

rivkahstandig3636:
@dozoisch Are you still experiencing this issue? And if so, what version are you on? This issue should have been fixed as of v2.3.0

dozoisch:
@rivkahstandig3636 I'll update and report back! I'm most likely still on 2.2.9

mathieumg:
@rivkahstandig3636 I work with @dozoisch and we're no longer experiencing the issue, sorry for not reporting back, and thanks for your work! 👍

rivkahstandig3636:
No problem, thanks @mathieumg!

msmaromi:
> Getting the same issue on the web. Seems like enabled: false with verbose: true doesn't log anymore. Will try to investigate later.

I actually still experience it

isaiah-coleman:
[...]

PR #473, opened 2017-10-23
Migrate from debug to binding console

guiconti:
As discussed in issue [#470](https://github.com/rollbar/rollbar.js/issues/470) this PR tries to keep verbose logging without the debug module


rokob:
Yeah I like this change, I don't think we should use debug anymore. 👍

PR #479, opened 2017-10-27
Fix how telemetry is reconfigured

rokob:
Correctly update the telemetry scrubbing settings in the browser instrumentation configure method


rokob:
:shipit:

Issue #482, opened 2017-11-01
Error: "Rollbar is not enabled"


simlu:
Why was this error introduced? Extremely confusing...

E.g. How are we supposed to get coverage in our tests now?

simlu:
This was changed in https://github.com/rollbar/rollbar.js/commit/a9864a43036e6357060172bbe3c2ce75100230c4 (Issue: https://github.com/rollbar/rollbar.js/issues/280)

simlu:
I've opened a pull request. Looking forward to some feedback....

rokob:
This does not throw an error, it calls your callback with an error if you call log and have Rollbar not enabled. This is the semantics required of the callback when we do not process a log call. It is not clear why this would have an effect on your tests.

simlu:
I don't use a callback. I'm using rollbar.wait instead... I think in that case it does throw an error? Let me double check where the error comes from.

simlu:
Yeah, this seems to be thrown from the rollbar.wait - urg

rokob:
The only time we throw from wait is in the lambda handler which is just a rethrow to allow for any other higher level error handling to still take place and to fail the request otherwise rather than swallowing errors.

simlu:
Removing `endpoint: "https://api.rollbar.com/api/1/",` from the options magically solved this. Wow, those were some bizarre errors. Apologies for not checking the change-log earlier.

rokob:
Sorry yeah, I wasn't happy about breaking that, but I needed it to get this library to match the semantics of rollbar-ios and rollbar-android so I could mix these together for React Native. I thought about going to 3.0.0 for it, but it seemed like bumping from 2.2.x to 2.3.0 should be enough.

simlu:
Makes perfect sense. React Native looks very promising. We're investigating it as well.

Suggestion: Possible to log a warning if somebody is using the old endpoint semantics? E.g. check with a regex? I wouldn't be surprised if other ppl run into this as well.

PR #483, opened 2017-11-01
Only error if enabled not explicitly set

simlu:
This would fix https://github.com/rollbar/rollbar.js/issues/482


simlu:
I misunderstood what this was doing. Removing endpoint: "https://api.rollbar.com/api/1/", from the options magically solved this. Those were some bizarre errors.

PR #492, opened 2017-11-04
fixes errMsg.match is not a function error

gofreddo:
I'm getting errMsg.match is not a function error from https://cdnjs.cloudflare.com/ajax/libs/rollbar.js/2.3.1/rollbar.min.js

### DIFF
--- a/src/browser/errorParser.js
+++ b/src/browser/errorParser.js
@@ -63,7 +63,7 @@ function parse(e) {
 
 
 function guessErrorClass(errMsg) {
-  if (!errMsg) {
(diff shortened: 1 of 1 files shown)


rokob:
Yeah this is a reasonable thing to check, I'd like to understand how we end up with an error with a message that does not have the String prototype. Unless someone throws an error that specifically sets message to null, because by default Error has the empty string as the message. But this is a good change, if we end up in that situation we shouldn't crash.

gofreddo:
Yeah I haven't been able to catch it.  Some library I'm using is obviously not throwing a well formed error.

PR #503, opened 2017-11-14
Be defensive when accessing event properties

### LINKED ISSUE #488: we're trying to handle rejected promise cross domain
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Property_access_denied

Error: Permission denied to access property "reason"

@rokob I know you're aware of this one, just created this for tracking purposes.

rokob:
Accessing properties of an event as part of an event handler on the window can violate the browser's SOP, so be careful

Fixes #488


papandreou:
Could we have a new release with this fix, pretty please? :)

rokob:
Yeah I will try to get one out today.

PR #510, opened 2017-11-20
Proper scrubbing

### LINKED ISSUE #508: Any field containing the consecutive letters "pw" gets scrubbed.
For example, if I have a property called `tempWorker` or `pickupWindow`, I wouldn't expect this property to get scrubbed, but it does.

This is made worse by the fact that ["pw" is not a documented default scrubbed field](https://github.com/rollbar/rollbar.js/issues/507), by the fact that seemingly default scrubbed fields cannot be overridden, only appended to (?), and _much, much worse_ by [the fact that scrubbing **mutates** logged objects](https://github.com/rollbar/rollbar.js/issues/509).

It seems like [this regular expression](https://github.com/rollbar/rollbar.js/blob/f0140a050c270981cfb980c9aa492a2d032fcb2f/src/utility.js#L550) should be matching on the whole string, i.e.

```js
'^\\[?(%5[bB])?' + scrubFields[i] + '\\[?(%5[bB])?\\]?(%5[dD])?$'
```

If that sounds reasonable, I can open a PR.

FWIW, this seems like this could also be a problem with "pass".

### LINKED ISSUE #509: Field scrubbing *mutates* logged objects!
The feature that scrubs fields detected to be "sensitive" – [which suffers from false positives](https://github.com/rollbar/rollbar.js/issues/508), is [poorly documented](https://github.com/rollbar/rollbar.js/issues/507), and seemingly cannot be disabled(?) – actually _mutates_ these properties on the logged objects. This is a pretty egregious breach of expected conventional node module behavior. If this is intentional (I hope not...), it should be very clearly indicated in the documentation.

The combination of this mutation bug and [the false positive bug](https://github.com/rollbar/rollbar.js/issues/508) made for a pretty nasty (though admittedly fascinating) bug, that I could easily see happening to others. We're using the [cookie-session](https://github.com/expressjs/cookie-session) middleware, which populates the contents of a user cookie on `req.session`. We attach a part of this user context to our logger, which ships errors to Rollbar with that context. The cookie-session middleware's behavior is to set a `set-cookie` header on the response when `req.session` has been changed. These bugs led to, when an error is logged, a boolean property named `tempWorker` being mutated to "********", which sent an updated corrupted cookie down to the browser, which was persisted then sent back on subsequent requests, which led to authentication errors. Oof.

Looks like [the mutation is in the `traverse` function](https://github.com/rollbar/rollbar.js/blob/f0140a050c270981cfb980c9aa492a2d032fcb2f/src/utility.js#L156) and it does seem like the behavior [was somewhat intentional](https://github.com/rollbar/rollbar.js/blob/f0140a050c270981cfb980c9aa492a2d032fcb2f/src/server/transforms.js#L175)...

I could open a PR if creating a deep clone of the object and returning a scrubbed copy in `scrub()` sounds good. Thanks!

rokob:
Fixes #508 
Fixes #509 

We should not modify an object in-place to scrub it.
We should not use an unanchored regular expression for testing whether to scrub a key.

### DIFF
--- a/src/utility.js
+++ b/src/utility.js
@@ -547,7 +547,7 @@ function _getScrubFieldRegexs(scrubFields) {
   var ret = [];
   var pat;
   for (var i = 0; i < scrubFields.length; ++i) {
-    pat = '\\[?(%5[bB])?' + scrubFields[i] + '\\[?(%5[bB])?\\]?(%5[dD])?';
+    pat = '^\\[?(%5[bB])?' + scrubFields[i] + '\\[?(%5[bB])?\\]?(%5[dD])?$';
     ret.push(new RegExp(pat, 'i'));
   }
   return ret;
--- a/src/utility.js
+++ b/src/utility.js
@@ -150,13 +150,14 @@ function traverse(obj, func, seen) {
     }
   }
 
+  var result = isObj ? {} : [];
   for (i = 0; i < keys.length; ++i) {
     k = keys[i];
     v = obj[k];
-    obj[k] = func(k, v, seen);
+    result[k] = func(k, v, seen);
   }
 
-  return obj;
+  return (keys.length != 0) ? result : obj;
 }
 
 function redact() {
@@ -539,8 +540,7 @@ function scrub(data, scrubFields) {
     }
   }
 
-  traverse(data, scrubber, []);
-  return data;
+  return traverse(data, scrubber, []);
 }
 
 function _getScrubFieldRegexs(scrubFields) {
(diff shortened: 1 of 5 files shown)


src/utility.js
```diff
@@ -150,13 +150,14 @@ function traverse(obj, func, seen) {
     }
   }
 
+  var result = isObj ? {} : [];
   for (i = 0; i < keys.length; ++i) {
     k = keys[i];
     v = obj[k];
-    obj[k] = func(k, v, seen);
+    result[k] = func(k, v, seen);
```
brianr: This seems like clearly the right thing to do, but just to ask... are there any places that we currently use `traverse` where we actually do/might want to modify in-place?

rokob: It is only used internally by the scrubbing method

Issue #519, opened 2017-12-06
Telemetry: ability to ignore certain network requests


brianr:
(Via customer request.)

It'd be nice to be able to ignore certain network requests from the telemetry timeline, e.g. ignore requests made to certain hosts.

Why:

- drop minimally-useful telemetry events so they don't clutter the UI
- avoid using up the whole telemetry queue (which we cap at a hard max of 100) with such events

msokk:
Would also love to have this. Our use case would be to filter out all mixpanel requests, as the each 4000 character payload is encoded into URL as base64. 

Blacklisting full url (with regexp) support would make sense or maybe even more generic callback based filter would make sense? `filterNetworkTelemetry(metadata): boolean`

msokk:
Mixpanel made up 80% of our telemetry content so I went ahead and created a PR #521

PR #521, opened 2017-12-07
Add filterNetworkTelemetry callback option

msokk:
Fix for #519.

Open for any naming and other suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,7 +412,15 @@ The different types of events that we automatically capture are: `network`, `log
 `navigation`, and `connectivity`.
 
 Network events are XHR and fetch requests. We store the status code, the url, and some timing events
-to determine how long requests take.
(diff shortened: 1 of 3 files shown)


rokob:
This seems reasonable to me. Looking at it I am trying to think whether we would just want a `filterTelemetry` option which operates the same way but is passed the entire object before adding it to the queue. It would go in `capture` and then to do this you would write

```js
{
  filterTelemetry: function(e) {
    return e.type === 'network'
      && (e.body.subtype === 'xhr' || e.body.subtype === 'fetch')
      && e.body.url.indexOf('https://spammer.com') === 0;
  }
}
```

It makes the filter function slightly more complicated and would not be worth it if people only ever want to filter network things. But it is a lot more flexible and achieves the same goal.

msokk:
Yup, you're right, it briefly passed through my mind, I'll open up another PR and close this

msokk:
Closing in favor of more generic solution #522

PR #522, opened 2017-12-08
Add filterTelemetry callback configuration parameter

### LINKED ISSUE #519: Telemetry: ability to ignore certain network requests

msokk:
Adds the capability to filter out any telemetry event with custom filter function.
Resolves #519.


msokk:
@rokob any thoughts?

PR #531, opened 2017-12-20
Fix Rollbar types

gonpombo8:
Rollbar version 2.3.3 

TypeScript

```js
import * as Rollbar from 'rollbar'
```
`Rollbar` does not have the methods `error/warning/info`, so calling it returns a type error.
` Property error does not exist on type 'typeof Rollbar'`

image

This exports all this methods as 'rollbar', so we can use it without calling Rollbar.init instance. 🎸

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -7,7 +7,7 @@
  *~ file should be 'super-greeter/index.d.ts'
  */
 
-const rollbar: Rollbar;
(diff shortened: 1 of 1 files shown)


rokob:
I am going to close this in favor of https://github.com/rollbar/rollbar.js/pull/537 which has more type updates and the changes here folded in.

PR #533, opened 2017-12-21
Fixed tiny React Native issues

michaldudek:
Two small fixes for React Native support:

- Removed `require('url')` as it's not used at all and is not available in React Native environment at all (another workaround is to `npm i url`), as opposed to nodejs.
- Fixed typo for `RegExp` as JS is case sensitive and this was throwing an `undefined` error.


rokob:
LGTM

Issue #486, opened 2017-11-01
rollbar.js floods with error


MaximKraev:
We are getting strange behaviour then we get a error occurrence reports about a error 60 times per minute and never stops, there is no way to stop it except blocking IP what we are currently doing,
In sept we had 23,649 error occurrences, now it is about 10 times that. Support can't help as and sends here, any idea how we can resolve that?

rokob:
Have you tried using the `maxItems` and `itemsPerMinute` settings?

MaximKraev:
this is not a solution, it just workaround we will be changed for anyway.

MaximKraev:
as I understand we hit exactly` itemsPerMinute` default rate and `maxItems` can filter real errors from client

rokob:
1. Those rate limits are on the client side, so you won't be charged for anything beyond those limits.
2. If there is an error in your code generating this many errors, you can use the `checkIgnore` function to filter out those specific errors.
3. If you believe this is a bug in rollbar.js then can you provide more details so I can debug?

MaximKraev:
1. Yes I know, but 60 items per minute is a limit currently, which is 86400 occurrences a day we are changed for
2. the error appears even before our code is loaded, it appears for different errors, some of them can't trigger such flood at all.
3. Could you please point me where I can get information you need?

MaximKraev:
Anything I can help you with to diagnose the issue?

simlu:
@rokob Which version does this happen with?

MaximKraev:
we are using 2.2.8 currently

MaximKraev:
we are using 2.2.8 currently

aharpervc:
What's the nature of the reports? I saw something like this, possibly due to unknown chrome extension behavior. I filter them out with `checkIgnore`:

```js
checkIgnore: function(isUncaught, args, payload) {
            // extremely conservative rejection for only the most obvious, high volume false positives
            return (isUncaught
                && args === undefined
                && payload.body.trace.exception.class == "(unknown)"
                && payload.body.trace.exception.description == "unhandled rejection was null or undefined!"
                && payload.body.trace.frames.length == 1
                && payload.body.trace.frames[0].filename == "(unknown)"
                && payload.body.trace.frames[0].method == "[anonymous]");
        },
```

MaximKraev:
we have filter for non site domains, also it happens for any errors. more often for "(unknown) Script error"

rokob:
[...]

PR #543, opened 2018-01-08
Send a more useful trace when reason=null

corps:
Related:
https://github.com/rollbar/rollbar.js/issues/486

From the rollbar unhandled rejection code paths, reason should, in the worst case, be a string, and not null.  It seems that something (extensions?  particular browser builds?  bad actors?) may be calling this internal method directly without the reason param.  This change ensures we send a full trace in that scenario.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ coverage
 .DS_Store
 *.swp
 npm-debug.log
+config.nix
+.provisioned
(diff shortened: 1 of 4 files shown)


corps:
@rokob After investigating a bit deeper, 
https://github.com/rollbar/rollbar.js/pull/503

You may have already started to fix this issue.  I'm guessing the reason may be null to deployments using code older than this pull.  Maybe this PR can be closed then?

PR #544, opened 2018-01-09
create cache to ignore duplicate exceptions

and-megan:
WIP

@corps and I were looking at this. To help with cases like https://github.com/rollbar/rollbar.js/issues/486. 

We're gonna refactor this to take it off the RateLimiter

### DIFF
--- a/src/rateLimiter.js
+++ b/src/rateLimiter.js
@@ -10,6 +10,7 @@ function RateLimiter(options) {
   this.platform = null;
   this.platformOptions = {};
   this.configureGlobal(options);
+  this.dupCache = {};
 }
 
 RateLimiter.globalSettings = {
@@ -36,6 +37,9 @@ RateLimiter.prototype.configureGlobal = function(options) {
   if (options.itemsPerMinute !== undefined) {
     RateLimiter.globalSettings.itemsPerMinute = options.itemsPerMinute;
   }
+  if (options.duplicateRateLimit !== undefined) {
+    RateLimiter.globalSettings.duplicateRateLimit = options.duplicateRateLimit;
+  }
 };
 
 /*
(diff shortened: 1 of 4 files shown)


src/rateLimiter.js
```diff
@@ -60,6 +64,28 @@ RateLimiter.prototype.shouldSend = function(item, now) {
...
+    //   message: message,
+    //   err: err,
+    //   custom: custom,
+    //   timestamp: now(),
+    //   callback: callback,
+    //   uuid: uuid4()
+    // };
+    var exceptionMessage = item.message;
+    var exception = item.err;
+    var exceptionHash = exceptionMessage + '-' + (exception.stack || String(exception));
```
rokob: Because we are checking `item._isUncaught` I am guessing that `item.err == exception` is therefore not undefined here. However, `item.message` would then always be undefined so that this line would throw. If message was not undefined then exception probably would be so that `exception.stack` would throw. This line needs some care, assume that everything could be undefined and just add some conditionals, something like

```
var exceptionHash = (exceptionMessage || '') + '-' + ((exception && (exception.stack || String(exception))) || '');
```

I'm not exactly sure which of those parentheses are superfluous if any

src/rateLimiter.js
```diff
@@ -60,6 +64,28 @@ RateLimiter.prototype.shouldSend = function(item, now) {
...
+
+    var lastOccurrenceMS = this.dupCache[exceptionHash];
+    this.dupCache[exceptionHash] = now;
```
rokob: The problem with this is that over time the memory footprint of `dupCache` will grow unbounded. That is not okay for server side js and it is even bad in some browser contexts as well. This is probably a scenario where an LRU cache is necessary instead of just using a js object as a cache.

Either write your own LRU cache using an object plus a doubly linked list, or I can do that, or we can pull something in from npm.

test/rateLimiter.test.js
```diff
@@ -51,6 +53,61 @@ describe('shouldSend', function() {
...
+    return { rollbar: rollbar, rateLimiter: rateLimiter, client: client };
+  };
+
+  var createItem = function(setup, message, exception) {
+    setup.rollbar.handleUncaughtException(message, '', 0, 0, exception, null);
+
+    return setup.client.logCalls[setup.client.logCalls.length - 1].item;
+  }
+
+  it.only('asdfasdf', function (done) {
```
rokob: I won't comment too much on this test as it appears to be a work in progress, but I would get rid of the magic numbers like `123123` and make that a constant like `now`, then the other ones can be offsets from there like `now + duplicateRateLimit - 1` and `now + duplicateRateLimit + 1`. `shouldSend` takes that second now parameter to make tests like this easy to write. But you lose the ability to use that by using the real current time in `setupDedupeTest` without getting a reference to that `now` value back to use in the tests.

corps:
This whole pull is a WIP.  Don't worry about reviewing for now.

rivkahstandig3636:
@corps @and-megan Any news on this PR or should @rokob take it over?

and-megan:
I think we're pretty heads down with other work. Thanks for checking @rivkahstandig3636. Sorry for forgetting about this!

rokob:
I'm going to close this, I already did 80% of the use case for this by getting the last error working based on a hash.

corps:
Seems reasonable.  This was honestly a fun little project me and megan picked up for 45 minutes :P It was fun to stab at the SDK.

PR #553, opened 2018-01-19
turn sourceMaps on for uglify

### LINKED ISSUE #491: No source maps published for the .min.js dist bundles?
As per https://github.com/rollbar/rollbar.js/blob/626aac317d7879a3ecf62b71b0cadfa0dcc226b4/package.json#L10 I'm using `dist/rollbar.umd.min.js` in production. I just ran into a scenario where an error occurred in rollbar itself, and I discovered that there's no `sourceMappingURL` ref in the file and no `rollbar.umd.min.js.map` either.

Seems like that goes for *.min.js in the dist dir: https://github.com/rollbar/rollbar.js/tree/master/dist

I'm assuming this is by accident?

rokob:
Fixes #491

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -7,10 +7,7 @@ var outputPath = './dist/';
 
 var defaultsPlugin = new webpack.DefinePlugin(defaults);
 var uglifyPlugin = new webpack.optimize.UglifyJsPlugin({
-  // We've had some reports of the sourceMappingURL comment causing problems in Firefox.
-  // The uglifyjs plugin doesn't provide a way to generate the source map without generating
-  // that comment, so until we can resolve that, let's just not generate the source map.
-  sourceMap: false,
+  sourceMap: true,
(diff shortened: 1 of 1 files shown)


papandreou:
@rokob, thanks for doing this! How about a new release with this change? 🙏

papandreou:
@rokob, ping? Pretty please?

rokob:
Ah yeah sorry will do.

rokob:
I released v2.3.8 which has this change

papandreou:
Thanks! Seems like the `//# sourceMappingURL=rollbar.umd.min.js.map` reference is missing from rollbar.umd.min.js, though? That makes it harder to get existing tools like https://github.com/webpack-contrib/source-map-loader to pick it up.

rokob:
🤦‍♂️  https://github.com/rollbar/rollbar.js/pull/568

PR #555, opened 2018-01-19
Update Angular integration guidelines

wozniakm:
Hey guys!

According to Angular's documentation, service locator pattern should be avoided unless there are no other options. In this case there are other options, we can inject `Rollbar` using `@Inject` decorator.

In former version, if you are not using `RollbarService` in different places of your application, `Rollbar` will be initialized only when error has happend. That's causing lack of telemetry information. `Rollbar` should be initialized on application initialization to collect all telemetry data.

To achieve that, we can inject it into root component of application or we can inject it into `ErrorHandler`, which is a part of Angular's core library and is initialized at the very beginning.

Would be awesome if you consider merging this change. It might help other Angular and Rollbar users.

Cheers!
Matt

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,7 @@ import * as Rollbar from 'rollbar';
 import { BrowserModule } from '@angular/platform-browser';
 import {
   Injectable,
+  Inject,
   Injector,
   InjectionToken,
   NgModule,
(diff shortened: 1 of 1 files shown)


rokob:
This looks good to me, thanks. We are admittedly not as experienced with Angular as we should be. I'm going to merge this. I'd be happy to add any other changes or advice to the README that you think would be helpful for Angular users.

PR #559, opened 2018-01-25
Adding in testing docs for Firefox

### LINKED ISSUE #458: Error in documentation: "TypeError: window.onerror is not a function"
In the documentation, to test things out, you write:
```
Type the following code into the console and press enter: window.onerror("TestRollbarError: testing window.onerror", window.location.href)
```
Doing this in Firefox raises a `TypeError: window.onerror is not a function`. This is because window.onerror is not accessible from the console.

In the end, I checked that the rollbar integration worked with this:
```
setTimeout(function() {notThere();}, 1000);
```
You could use a snippet like this in your documentation :)

rivkahstandig3636:
Fixes #458

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ If you're running Rollbar on an environment besides production, change the `envi
 1. Navigate your browser to a page that has the above code installed
 2. Type the following code into the console and press enter: `window.onerror("TestRollbarError: testing window.onerror", window.location.href)`
 
+If you're on Firefox, `window.onerror` is not accessible from the console, so you'll need to try this in your code instead: `setTimeout(function() {notThere();}, 1000);`.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -36,6 +36,8 @@ If you're running Rollbar on an environment besides production, change the `envi
 1. Navigate your browser to a page that has the above code installed
 2. Type the following code into the console and press enter: `window.onerror("TestRollbarError: testing window.onerror", window.location.href)`
 
+If you're on Firefox, `window.onerror` is not accessible from the console, so you'll need to try this in your code instead: `setTimeout(function() {notThere();}, 1000);`.
```
brianr: Nice - should we just suggest that everyone does this regardless of browser?

PR #577, opened 2018-02-15
Make sure context becomes a string if it exists and is not already one

### LINKED ISSUE #576: API error: Invalid format. data.context object value found, but a null is required.
Unless we specify a `payload.context` in the rollbar config we get the following bug. My question is what is causing this error—having trouble locating it in the code—and how might we fix it by altering our config or modifying the error middleware. It seems like the form of the error is missing something rollbar is expecting. Using v2.3.8.

**App setup**
We have a node app that uses express 
```javascript
const rollbar = new Rollbar({
  accessToken: process.env.POST_SERVER_ITEM_ACCESS_TOKEN,
  captureUncaught: true,
  captureUnhandledRejections: true,
  payload: {
    environment: process.env.NODE_ENV,
    context: 'somestring', // This line is required to avoid the errors
    server: {
      host: hostname,
      branch
    }
  }
});

```
This configuration also produces the bug
```javascript
const rollbar = new Rollbar(process.env.POST_SERVER_ITEM_ACCESS_TOKEN);
```
Then the express middleware with rollbar as error handler:
 ```javascript
app.use(rollbar.errorHandler());
```

**Reproducing**
try to get a route
```GET /adsf 500 43.132 ms - -```

Then deliberately throw an error in an express middleware. 🤢

**Expected behavior**
get the standard log with a rollbar URL to view on rollbar.com

**Actual behavior**
Upload to rollbar fails.
```GET /adsf 500 43.132 ms - -
Received error: Invalid format. data.context object value found, but a null is required.
Error reporting to rollbar, ignoring: Error: Api error: Invalid format. data.context object value found, but a null is required.
Error: only http(s) protocols are supported
```
or
image

rokob:
Fixes #576

### DIFF
--- a/src/apiUtility.js
+++ b/src/apiUtility.js
@@ -1,7 +1,7 @@
 var _ = require('./utility');
 
 function buildPayload(accessToken, data, jsonBackup) {
-  if (_.isType(data.context, 'object')) {
+  if (!_.isType(data.context, 'string')) {
     var contextResult = _.stringify(data.context, jsonBackup);
     if (contextResult.error) {
       data.context = 'Error: could not serialize \'context\'';


johnelliott:
Thanks @rokob 👍

Issue #578, opened 2018-02-16
input needs to be null checked


johnnywang:
https://github.com/rollbar/rollbar.js/blob/58fda62c63e9823a3148b3ece665d0306a880dec/dist/rollbar.noconflict.umd.js#L4720 is reading `url` from `input`, which is sometimes null, causing exceptions like this to get logged to Rollbar: `undefined is not an object`

PR #579, opened 2018-02-17
More telemetry options

rokob:
Introduce the `includeItemsInTelemetry` option to allow turning on/off the automatic inclusion of previous items sent to Rollbar. We make this default to false on the server side and true everywhere else.


rokob:
Also Fixes #578

PR #581, opened 2018-02-17
make the verbose option turn on/off all logging

### LINKED ISSUE #574: Turning off console output on the server.
I don't want to see the `Successful api response.` messages every time an error is logged.

It is documented that the debug package is used and the output can be turned off with the `verbose` option. Looking at the source, I see that `server/logger` outputs directly to console without any way to turn it off (older version _did_ use debug). Could this be implemented? Thanks!

rokob:
Fixes #574


rokob:
It is false by default even if the code inside the logger looks the other way because of the default configuration options. I’ll change the one inside the logger so that is less confusing.

PR #627, opened 2018-06-20
Scrub values of some known DOM elements

### LINKED ISSUE #625: Security | scrubTelemetryInputs default off
Hi guys,
We just notice that you not only introduce Telemetry with default settings on on our sites, also you turn off as default scrubTelemetryInputs.
We are sad :(

rokob:
By default we capture the input values of DOM elements in the telemetry data. This PR adds a set of patterns to scrubFields to test the `name` value of the DOM elements against. If they match we scrub the value. The list of strings comes from:

https://stackoverflow.com/questions/25163891/when-is-the-scan-credit-card-option-available-on-ios8-safari/36151606#36151606
https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#inappropriate-for-the-control

If `scrubTelemetryInputs` or `telemetryScrubber` is set then we do not do anything. By default `scrubTelemetryInputs` is false (this being true means scrub ALL values not just some possibly sensitive ones), and `telemetryScrubber` is a function that passes the description of each element to it and asks you to decide whether to capture the value or not. By default `telemetryScrubber` is undefined. So the assumption being that if you choose to define `telemetryScrubber` then you have full control over what is captured so it is up to you to decide what to do and we don't want to step on that if you really do want to capture some things that we would otherwise have scrubbed by default.

Fixes #625


rivkahstandig3636:
Thanks @rokob! Can you make sure to update the docs when this is released?

PR #645, opened 2018-07-17
Properly call the existing onerror handler if set

### LINKED ISSUE #644: Error handling fallback when Rollbar isn't loaded properly
On the application I work on, when an unhandled error occurs it gets sent to Rollbar and a modal is displayed  indicating that an error occurred and forces the user to refresh the page.

I am seeing that for some unknown reasons *rollbar.min.js* is not being loaded in some situations. This causes the `onSendCallback` to never trigger.

It would be ideal if we can also override `window.onerror` with our own logic that gets triggered if Rollbar is not loaded correctly.

Something like the following.

```javascript
var rollbarOnerror = window.onerror
window.onerror = function() {
  rollbarOnerror(/* arguments? */);
  if (!window.handledByRollbar) {
    // Perform some kind of backup, i.e. as user to send an email and force to refresh
  }
}
```

```javascript
// rollbar configuration

{
  // ...
  onSendCallback: function(uncaught) {
    window.handledByRollbar = true
    // ...
  }
  // ...
}
```

This can be artificially reproduced by blocking the *rollbar.min.js* request (in devtools for example). In my specific example I'm blocking https://cdnjs.cloudflare.com/ajax/libs/rollbar.js/2.3.9/rollbar.min.js (to try an create a fix for this situation actually occurring).

rokob:
Fixes #644

### DIFF
--- a/src/browser/globalSetup.js
+++ b/src/browser/globalSetup.js
@@ -4,8 +4,11 @@ function captureUncaughtExceptions(window, handler, shim) {
 
   if (typeof handler._rollbarOldOnError === 'function') {
     oldOnError = handler._rollbarOldOnError;
-  } else if (window.onerror && !window.onerror.belongsToShim) {
+  } else if (window.onerror) {
     oldOnError = window.onerror;
+    while (oldOnError._rollbarOldOnError) {
+      oldOnError = oldOnError._rollbarOldOnError;
+    }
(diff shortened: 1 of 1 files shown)


uri:
I tested this by adding the snippet from this branch; it behaves how I expect it to now. 👍

PR #656, opened 2018-08-10
Added support for tracking people on ExpressJS using user.sub instead of user.id

Ricardonacif:
Hey guys,

I just patched this. The req.user.sub is very common these days due to JSON Web Token, so I think it also makes sense that rollbar tries to identify the user by the sub attribute if the id attribute is not present.

Thanks,

### DIFF
--- a/src/server/transforms.js
+++ b/src/server/transforms.js
@@ -139,7 +139,7 @@ function addRequestData(item, options, callback) {
     }
     item.data.person = person;
   } else if (req.user) {
-    item.data.person = {id: req.user.id};
+    item.data.person = {id: req.user.id || req.user.sub};
     if (req.user.username && captureUsername) {
       item.data.person.username = req.user.username;
     }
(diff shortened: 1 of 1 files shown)


rivkahstandig3636:
Thanks for the PR! We'll check it out :)

rivkahstandig3636:
Hi @Ricardonacif, it looks like you would be able to add this information using the transform function (docs on this can be found [here](https://docs.rollbar.com/docs/javascript#section-transforming-the-payload)) so I don't think we'll be adding this to the SDK at this time. Thanks for the PR!

shakiba:
@Ricardonacif Please take a look at the same function that you changed, it is a transformer. It seems that `var req = item.request` would work.

Ricardonacif:
Yeah I tried item.request there but is null. Here's my initialization, am I doing something wrong?
https://gist.github.com/Ricardonacif/182a230d58867e645439fd84d715c50b

PR #657, opened 2018-08-24
fix(setPerson): configure notifier and telemeter with payload params

ninjz:
People Tracking does not work on the rollbar-react-native project. As talked about in [this](https://github.com/rollbar/rollbar-react-native/issues/39) closed issue. I have tracked down the issue to this JS library, where the configure method does not merge in the configured payload data into the notifier and telemeter.

### DIFF
--- a/src/rollbar.js
+++ b/src/rollbar.js
@@ -36,13 +36,13 @@ Rollbar.prototype.global = function(options) {
 };
 
 Rollbar.prototype.configure = function(options, payloadData) {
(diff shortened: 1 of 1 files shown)


shakiba:
@ninjz Thanks for PR! What about moving it to after `this.options = ...` and pass `this.options`.

ninjz:
@shakiba Thanks for the review, I've made those requested changes.

shakiba:
Thank you @ninjz !

Issue #670, opened 2018-09-12
ObjectID not stringified since v2.3.2


arlo-egg:
Description:
When logging the properties of a mongo object it is often useful to know the id of the object (which has a type of [`ObjectID`](https://docs.mongodb.com/manual/reference/method/ObjectId/)). Since v2.3.2 was released the id is treated as an object when logged and expanded out to constituent parts (the underlying buffer). This is much less useful for looking up the object in question because it has to be converted back to a string before using in a query.

I have tested in v2.4.5 and this bug still exists. I assume it is a bug because the change in behaviour is not mentioned in the [changelog](https://github.com/rollbar/rollbar.js/blob/master/CHANGELOG.md#v232).

Repro:
```
const ObjectID = require('mongodb').ObjectID
Rollbar.error("this is a test", {id: new ObjectID()})
```

Output in v2.3.1 (desired)
`id: 5b984b399c8bf914c75f6084 `

Output since v2.3.2 (actual)
```
id._bsontype: ObjectID
id.id.data: [91, 152, 72, 63, 210, 86, 64, 10, 29, 53, 180, 88]
id.id.type: Buffer
```

jessewgibbs:
Hey Arlo,

Thanks for the detailed bug report.  We'll investigate what changed since 2.3.1 and follow up.

shakiba:
@arlo-egg Thanks for reporting the issue, and also for very helpful details. I released v2.4.7 with a fix for this issue.

jessewgibbs:
Hi @arlo-egg ,

This issue is fixed in the 2.4.7 release.  Can you give that a try and let us know if it's working as expected?

arlo-egg:
Hi @jessewgibbs, @shakiba I have upgraded and verified that it is logging ObjectIDs as expected. Thanks for the fix!

PR #674, opened 2018-09-20
use partial cloning in scrubber

shakiba:
In scrubber, changed cloning to partial cloning (similar to immutable/persistent data structure), to avoid changing how unscrubbed data are stringified when send to server. See #670.


shakiba:
@rokob If possible I recommend using second parameter of JSON.stringify for scrubbing sensitive data instead, I mean when item data is being prepared for transport.

PR #687, opened 2018-10-25
Use the constructor name on the error if there is one

### LINKED ISSUE #686: custom error class is ignored
Rollbar 2.4.6

```
class MyError extends Error {}
Rollbar.error(new MyError("my message"))
```

Payload has `exception.class` set to `Error` instead of `MyError`

image

rokob:
Fixes #686

### DIFF
--- a/src/browser/errorParser.js
+++ b/src/browser/errorParser.js
@@ -61,7 +61,9 @@ function Stack(exception) {
   }
 
   var name = exception.constructor && exception.constructor.name;
-  name = name || exception.name;
+  if (!name || !name.length || name.length < 3) {
+    name = exception.name;
+  }
 
   return {
     stack: getStack(),
--- a/src/browser/errorParser.js
+++ b/src/browser/errorParser.js
@@ -60,10 +60,13 @@ function Stack(exception) {
     return stack;
   }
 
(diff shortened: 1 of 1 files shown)


agriffis:
This can still be foiled by uglifyjs renaming functions. It won't get the _wrong_ name (like "Error"), it'll just get a _meaningless_ name (like "e")

One approach is to modify uglifyjs config as shown in #686. That's complicated though... and not available in the popular create-react-app (where the webpack config can't be modified without forking)

Another approach is to set the name explicitly in the constructor. The caveat is that you can only do this on the instance, you can't explicitly set the name of the constructor (because function names are read-only once this is transpiled):

```
class MyError extends Error {
  constructor(...args) {
    super(...args)
    this.name = 'MyError'
  }
}
```

In that case, you'd want to prioritize `exception.name` over `exception.constructor.name`

Is this too clever?

```
var name =
  exception.constructor &&
  exception.constructor.name &&
  exception.constructor.name.length > 2
    ? exception.constructor.name
    : exception.name;
```

jessewgibbs:
@rokob can this be merged and released?

rokob:
I’m going to update this to the suggest check which I think isn’t a bad option considering what we have to work with. Then I’ll merge it.

Issue #549, opened 2018-01-15
WebWorker support


solgarius:
I am having trouble using the module inside a webworker with webpack build (though I suspect it would fail via other code bundling tools). 

I am getting the error:
```
Uncaught ReferenceError: window is not defined
    at Object.<anonymous> (rollbar.umd.min.js:1)
    at e (rollbar.umd.min.js:1)
    at Object.<anonymous> (rollbar.umd.min.js:1)
    at e (rollbar.umd.min.js:1)
    at rollbar.umd.min.js:1
    at rollbar.umd.min.js:1
    at rollbar.umd.min.js:1
    at Object.<anonymous> (rollbar.umd.min.js:1)
    at __webpack_require__ (bootstrap 12d908e1a57503409ef5:19)
...
```
As webworkers don't have the `window` global and rather use the `self` global the code would need to check for existence of each to use appropriately.

I am using the rollbar.umd.js dist however all dists have this same issue. The offending code lies in cases where testing for the window variable e.g.
```
	var options = window && window._rollbarConfig;
```
On chrome this throws the above error and the existence of window should instead be checked by:
```
	var options = (typeof window !== 'undefined') && window._rollbarConfig;
```
For proper support would probably need a function that resolves the proper global to use e.g.
```
function getGlobal(){
  if(typeof window !== 'undefined'){
    return window // browser main renderer
  } else if(typeof self !== 'undefined'){
    return self // web worker
  }
  return null 
}
```
Might also want to do a check for `global` as a 3rd option for app build tools like electron, however electron processes may want to use the server version of rollbar.

Additionally for web worker support, might need to submit the error in a different way. I didn't check whether the rollbar module uses a method to submit errors to rollbar.com that is supported in web workers.

rivkahstandig3636:
Thanks for the report! We'll take a look.

PR #703, opened 2019-03-03
Support web workers by passing gWindow instead of window

pbiggar:
In #549, support was added for using `self` in web workers, instead of `window`. This was called `gWindow`. However, `addTransformsToNotifier` continued to use `window`. 

Before this, if I called `Rollbar.init` inside a web worker, it would throw an exception. Now it does not.

I tried to rebuild using `npm run-script build` and I got an unrelated-seeming error:

```
ERROR in ./src/utility.js
Module not found: Error: Cannot resolve 'file' or 'directory' ../vendor/JSON-js/json3.js in /Users/paulbiggar/projects/rollbar.js/src
 @ ./src/utility.js 22:26-63
Warning: Task "webpack:2" failed. Use --force to continue.
```

Also, the generated files had lots of unrelated changes, so I didn't include them. Presumably there's some process that updates them.

### DIFF
--- a/src/browser/rollbar.js
+++ b/src/browser/rollbar.js
@@ -21,7 +21,7 @@ function Rollbar(options, client) {
 
   var gWindow = ((typeof window != 'undefined') && window) || ((typeof self != 'undefined') && self);
   var gDocument = (typeof document != 'undefined') && document;
-  addTransformsToNotifier(this.client.notifier);
+  addTransformsToNotifier(this.client.notifier, gWindow);
   addPredicatesToQueue(this.client.queue);
   if (this.options.captureUncaught || this.options.handleUncaughtExceptions) {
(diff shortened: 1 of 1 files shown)


waltjones:
@pbiggar, thank you for the PR. That error is resolved by running `git submodule update --init`. Definitely unrelated to this PR.

This looks good to me. @rokob any comments?

rokob:
Ah yes thank you, I'm not sure how I missed those window calls. 👍

pbiggar:
Thanks! Do you have a release schedule for this to be in npm?

rokob:
I just went through the backlog to make sure there isn't something small that I wanted to get out quickly as well. There isn't, so I will just push a release now.

pbiggar:
Oh wonderful, thanks!

Issue #659, opened 2018-08-27
Telemetry Capture Event Type


ajtran:
internal: https://app.intercom.io/a/apps/rlisg82q/inbox/inbox/488687/conversations/18151807899

User would like to specify what type of event is being captured when using the captureEvent method. As of now, the type is always being set to `manual`.

PR #704, opened 2019-03-06
Allow telemetry captureEvent to specify optional event type

waltjones:
Fixes https://github.com/rollbar/rollbar.js/issues/659

Allows `captureEvent()` to be used as before with default event type `'manual'` or to specify an event type, by allowing flexible argument order similar to the way `log()` works.

Example:
```
captureEvent({foo: 'bar}, 'info'); // defaults to 'manual' event type

captureEvent('log', {message: 'bar}, 'info'); // Flexible argument ordering allows event type to be first argument
```

### DIFF
--- a/src/utility.js
+++ b/src/utility.js
@@ -236,9 +236,8 @@ function parseUri(str) {
   var o = parseUriOptions;
   var m = o.parser[o.strictMode ? 'strict' : 'loose'].exec(str);
   var uri = {};
-  var i = o.key.length;
 
-  while (i--) {
+  for (var i = 0, l = o.key.length; i < l; ++i) {
     uri[o.key[i]] = m[i] || '';
   }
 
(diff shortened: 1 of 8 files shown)


src/utility.js
```diff
@@ -453,6 +453,55 @@ function createItem(args, logger, notifier, requestKeys, lambdaContext) {
...
 }
 
+var TELEMETRY_TYPES = ['log', 'network', 'dom', 'navigation', 'error', 'manual'];
+var TELEMETRY_LEVELS = ['critical', 'error', 'warning', 'info', 'debug'];
+
+function arrayIncludes(arr, val) {
+  var len = arr.length;
+  var k = 0;
+
+  while (k < len) {
```
rokob: Is there a reason for a while loop where and not a for loop?

waltjones: No, it's arbitrary. Is there a reason to prefer a for loop?

rokob: Not really, just consistency. This would be only the second while loop in this file and the only other one in `parseUri` could use a cleanup. I mean they are the exact same when it comes to execution, but I was just surprised to see a while loop in this function and a for loop in the next function within the same PR.

rokob: Same with `++i` vs `i++`, in the loop postamble they are the same but throughout this codebase we mix them a bit, I'd other stick with one. I can go through after this PR and clean up all my weird style nits.

waltjones: All makes sense. I'll take care of these before merge.

waltjones: I went ahead and also cleaned up `parseUri`, and converted the few postfix increment/decrements in utility.js to prefix.

PR #706, opened 2019-03-08
chore: Add wait type def

### LINKED ISSUE #705: Rollbar wait(...) type definition
When trying to use rollbar with typescript, I noticed that there is no method for `wait(...)` in the type definition file. 

Looks like a pretty easy one to add. Should just be

```js
...
public wait(callback: () => void): void;
...
```

ryanburr:
Closes #705


rokob:
LGTM

PR #709, opened 2019-03-14
If we already handled an error inside a wrap, don't do it again

rokob:


rokob:
There are scenarios where we end up with multiple layers of functions being wrapped, primarily across instance boundaries. We want to preserve the deepest context so we only want to use the global error cache business from the deepest level. This change makes us skip reprocessing the same error.

PR #711, opened 2019-03-19
Improved LambdaHandler event typing

vorks:
`@types/aws-lambda` allows the event to be typed as shown here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/index.d.ts#L1028

Wrapping the handler with `lambdaHandler()` removed the `@types/aws-lambda` typing. Forcing the `event` to always be of type `object` breaks the expected type of `event` in the handler. This change would enable you to type `lambdaHandler()` similarly to `Handler` from `@types/aws-lambda`. It would reinstate the desired type of `event` for the handler.

It should also be considered to enable typing of the result and to use the `Context` type from `@types/aws-lambda` instead of `object`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -21,7 +21,7 @@ declare class Rollbar {
 
     public captureEvent(metadata: object, level: Rollbar.Level): Rollbar.TelemetryEvent;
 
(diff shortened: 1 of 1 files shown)


waltjones:
@vorks Thanks your for the PR, and the detailed explanation.

PR #721, opened 2019-04-11
Enable RollbarProxy for sending Rollbar events

waltjones:
Cordova support is driving the need for this, but it can be used by anyone who makes a `RollbarProxy` object available.

### DIFF
--- a/src/browser/transport.js
+++ b/src/browser/transport.js
@@ -71,7 +71,23 @@ function _makeZoneRequest(accessToken, url, method, data, callback, requestFacto
   }
 }
 
+/* global RollbarProxy */
+function _proxyRequest(json, callback) {
+  var rollbarProxy = new RollbarProxy();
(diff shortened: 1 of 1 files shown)


src/browser/transport.js
```diff
@@ -71,7 +71,23 @@ function _makeZoneRequest(accessToken, url, method, data, callback, requestFacto
   }
 }
 
+/* global RollbarProxy */
+function _proxyRequest(json, callback) {
+  var rollbarProxy = new RollbarProxy();
```
ArturMoczulski: I understand that the idea behind `RollbarProxy` is to come from other JS sources, just like in the Cordova SDK. Is that correct?

ArturMoczulski:
LGTM :)

src/browser/transport.js
waltjones: Right. `RollbarProxy` is defined externally.

Issue #723, opened 2019-04-15
Add public interface to serialize payloads and send serialized payloads


brianr:
Motivation is this use case from React Native / Expo: there is a need to write the payload (in javascript) to a disk, then send it later on restart

To accomplish this, rollbar.js can provide:

- a way to generate the final payload and return it as a string (i.e. the serialized json payload) - i.e. `rollbar.serializePayload(payload)`
- a way to take a serialized string payload and send it over the network to api.rollbar.com - i.e. `rollbar.sendPayloadStr(payloadStr)`

jjabba:
Here is the related expo documentation, explaining how [expo deals with errors in production](https://docs.expo.io/versions/latest/guides/errors/). 

Basically, when an error occurs you can intercept it, and quickly store relevant data using `ErrorRecovery.setRecoveryProps(yourRelevantData)`. As explained [here](https://docs.expo.io/versions/latest/sdk/error-recovery/).

Once expo has killed and restarted your app, it will be able to access the data again via a prop `exp.errorRecovery` which is passed to your app. This is where it would make sense to pass the data on to rollbar via an asynchronies call to the rollbar.error() function.

waltjones:
The referenced PR enables this for rollbar.js and any use case that uses rollbar.js directly.

For users of rollbar-react-native, an update is needed in that repo because it wraps both the config and the public endpoints for rollbar.js.

@jjabba I'm not familiar enough with Expo to know which way you might be including Rollbar.

jjabba:
Thanks @waltjones! While I am using `rollbar-react-native`, I'm not actually using the wrapped Client, as I'm unable to link the native rollbar client while still using expo. I get around the initialization by importing and using `import Client from 'rollbar/src/react-native/rollbar'`

I've already prepared the `checkIgnore` hook in my configuration, looking forward to the PR making it into an actual release. 

What is the release cadence?

jjabba:
While I am using rollbar-react-native, I'm not actually using the wrapped Client, as I'm unable to link the native rollbar client while still using expo. I get around the initialization by importing and using import Client from 'rollbar/src/react-native/rollbar'

waltjones:
@jjabba good to know. Thank you!

I'll work on getting a release out for you asap.

waltjones:
@jjabba this is now published in v2.6.0.

brianr:
[...]

PR #724, opened 2019-04-19
Add public interface to serialize payloads and send serialized payloads

waltjones:
Fixes: https://github.com/rollbar/rollbar.js/issues/723

Adds two new API endpoints, allowing Rollbar event payloads to be captured as JSON, and delivered later. One example use case involves storing an exception event so it can be sent after app restart.

`buildJsonPayload()` accepts a `payload` object as provided in `checkIgnore()` for example, and returns a fully composed JSON string that can be stored or sent to the API as is.

`sendJsonPayload()` accepts the above string or any valid, fully prepared JSON string and sends to Rollbar API. 

The example (added to the webpack example) uses `checkIgnore()` to receive the `payload` object and build the JSON. This is a recommended place to hook this functionality because by returning `true` you can  prevent sending the stored payload.


jjabba:
Thanks @waltjones! While I am using rollbar-react-native, I'm not actually using the wrapped Client, as I'm unable to link the native rollbar client while still using expo. I get around the initialization by importing and using import Client from 'rollbar/src/react-native/rollbar'

I've already prepared the checkIgnore hook in my configuration, looking forward to the PR making it into an actual release.

What is the release cadence?

waltjones:
@jjabba this is now published in v2.6.0.

jjabba:
@waltjones, that's great! I just upgraded to v2.6.0. Serialization works great. I'm able to save the payload in the checkIgnore hook. However, Im running into a problem while trying to use `rollbar.sendJsonPayload(serializedError)`. The error I'm getting is:
     
     this.transport.postJsonPayload is not a function (In 'this.transport.postJsonPayload(this.accessToken,n,t,s)', 'this.transport.postJsonPayload' is undefined)

I looked into it, and it seems rollbar-react-native comes with it's own version of the transport implementation: `node_modules/rollbar/src/react-native/transport.js` and it doesn't seem to include the postJsonPayload function in it's export

     module.exports = {
       get: get,
       post: post
     };



 The corresponding browserfile `node_modules/rollbar/src/browser/transport.js` however, does :

     module.exports = {
       get: get,
       post: post,
       postJsonPayload: postJsonPayload
     };

waltjones:
@jjabba Sorry, I didn't see your comment until now. I will follow up shortly with another PR for this.

waltjones:
@jjabba https://github.com/rollbar/rollbar.js/pull/725 is in release 2.6.1. Apologies for the inconvenience. Let me know if you have any issues.

Issue #726, opened 2019-04-25
Setting captureUncaught and captureUnhandledRejections in configure has no effect


astorije:
We currently have a shared library where we define our Rollbar integration:

```ts
class Notifier extends Rollbar {
  constructor() {
    const { enabled, verbose, clientAccessToken, appVersion } = config.rollbar;

    super({
      enabled: true,
      accessToken: 'xyz',
      payload: {
        environment: process.env.NODE_ENV,
        client: { javascript: { code_version: 'abc' } },
      },
    });
  }

  captureUnhandledErrors() {
    this.configure({
      captureUncaught: true,
      captureUnhandledRejections: true,
    });
  }
}

export const notifier = new Notifier();
```

In our consumer applications, the first thing we do is this:

```
import { notifier } from 'whatever';
notifier.captureUnhandledErrors();
```

(This is to keep our library fairly side-effect-free, and make uncaught error capture intentional)

However, doing this does not send uncaught errors to Rollbar.
When moving `captureUncaught: true, captureUnhandledRejections: true,` to the constructor, uncaught errors are sent to Rollbar as expected.

Are we doing something wrong?
It seems to me that this is caused by [this](https://github.com/rollbar/rollbar.js/blob/bee040abee2cfafb0c3893c048e2889099ab5975/src/browser/rollbar.js#L26-L32): setting these options in the constructor calls other functions, but setting them in `configure` does not. I haven't dug too deep so I'm happy to be wrong.

waltjones:
You aren't doing it wrong, and you are correct that it doesn't currently work. It looks like setting them false later doesn't disable exception capture either.

This looks like something that could probably get into the next release. (~2 weeks?) I'll update here.

astorije:
Hey @waltjones, do you happen to have any update on this? :)
Thanks in advance!

waltjones:
@astorije In progress. Pull request soon. Release is probably Monday - Tuesday.

waltjones:
@astorije This is in v2.7.0.

PR #729, opened 2019-05-11
Allow enable/disable uncaught errors/rejections in .configure

waltjones:
Fixes https://github.com/rollbar/rollbar.js/issues/726

Previously, updating `captureUncught` or `captureUnhandledRejections` in `Rollbar.configure` would have no effect. This PR allows both enabling and disabling via `configure` in both the browser and server/node environments.

This PR also adds browser and server tests for unhandled error and rejection behavior, as well as tests for enabling/disabling via the constructor and configure.

### DIFF
--- a/examples/error.html
+++ b/examples/error.html
@@ -1,0 +1,23 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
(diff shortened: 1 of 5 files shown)


ArturMoczulski:
lgtm 👍 great work

PR #730, opened 2019-05-14
Enable rate limiting for node/server

waltjones:
Uses response headers from the Rollbar API to avoid transmitting while rate limiting is in effect.

If an API response returns statusCode: 429, and 'x-rate-limit-remaining': 0, API requests will be dropped until x-rate-limit-remaining-seconds have elapsed.

Because the user may change rate limit settings in the Rollbar dashboard, it has been requested to poll every 60 seconds to see if the rate limit status has changed. This implementation sets its remaining seconds to a maximum of 60. This accomplishes the same end as continuous polling, but only attempts to send when there is an actual event to transmit.

### DIFF
--- a/src/server/rollbar.js
+++ b/src/server/rollbar.js
@@ -7,7 +7,7 @@ var _ = require('../utility');
 var API = require('../api');
 var logger = require('./logger');
 
-var transport = require('./transport');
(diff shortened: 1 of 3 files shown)


ArturMoczulski:
great work 👍

PR #736, opened 2019-06-03
Add transmit option flag

waltjones:
To support SDK dev options, adds the `transmit` option, default `true`. Set to false to fully process events, but not send to Rollbar.

### DIFF
--- a/src/queue.js
+++ b/src/queue.js
@@ -89,6 +89,7 @@ Queue.prototype.addItem = function(item, callback, originalError, originalItem) 
   this._maybeLog(item, originalError);
   this.removePendingItem(originalItem);
   if (!this.options.transmit) {
+    callback(new Error('Transmit disabled'));
     return;
   }
   this.pendingRequests.push(item);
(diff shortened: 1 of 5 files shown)


src/queue.js
```diff
@@ -88,6 +88,9 @@ Queue.prototype.addItem = function(item, callback, originalError, originalItem)
   }
   this._maybeLog(item, originalError);
   this.removePendingItem(originalItem);
+  if (!this.options.transmit) {
```
brianr: - This might be the right place, or perhaps this should go in `_makeApiRequest` - what do you think?
- Looks like we are no longer calling the callback - is that intentional?

waltjones: This location is (I believe) after everything the caller would want to execute without sending the request. At entry of `_makeApiRequest` is logically equivalent. Any later would engage rate limiting. (And maybe that's desired? I'm not sure, but I thought not.)

I skipped the callback, since this case neither has a response nor produces an error. But it probably needs to be called in order for async/Promise wrappers to work correctly. I imagine other situations as well. I'll add it.

PR #744, opened 2019-06-28
Enhanced error report for anonymous errors

waltjones:
On Chrome, when onerror has a null error object, uses `Error.prepareStackTrace` to send more complete error information. This allows Chrome to show the same level of detail as Firefox for these errors.

Feature is enabled by default. Set `config.inspectAnonymousErrors = false` to disable.


ArturMoczulski:
Looks great. I'd love to see the follow up PR with postponed bubbling up of the error as well as discussed earlier today :)

PR #753, opened 2019-07-24
Add configured options to payload notifier object

waltjones:
For internal support team diagnosis, stores the user configured options in the payload notifier object.


ArturMoczulski:
Looks great @waltjones One thing I can't figure out is if you are attaching the full dump of all the config options or a delta between defaults and custom

ArturMoczulski:
Also, can you verify that the configured options in the payload under notifier key are being accepted by the api? It’s a new payload field and I’m not sure how strict the api is on accepting those

PR #755, opened 2019-07-29
Add ignoreDuplicateErrors config flag

waltjones:
Rollbar.js discards consecutive identical uncaught errors after sending the first error. This new flag allows overriding that behavior and transmitting all duplicate errors.

### DIFF
--- a/src/browser/rollbar.js
+++ b/src/browser/rollbar.js
@@ -529,7 +529,8 @@ var defaultOptions = {
   sendConfig: false,
   includeItemsInTelemetry: true,
   captureIp: true,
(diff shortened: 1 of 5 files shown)


brianr:
For consideration: would this flag make more sense in reverse? i.e. the feature here is to detect duplicates and then *not* transmit them. So maybe the flag could be `dropDuplicateErrors` or `ignoreDuplicateErrors`, with a default value of `true`

waltjones:
@brianr The more I look at it, `ignoreDuplicateErrors` seems better.

PR #758, opened 2019-07-31
Fix browser detection inside web workers

rianmcguire:
Web workers don't have document at all, so this was failing with "document is not defined".


waltjones:
@rianmcguire Thank you!

Issue #801, opened 2019-11-22
Can't run linter or tests following instructions in Developing


kevinrobinson:
hello!  I found some scrubbing behavior that didn't work the way I expected.  I forked and make the change, then tried to run the linter and tests, but have been stymied so far :)

### tests
From the contributing docs:
```
git submodule update --init
npm install -D
make
make test
````
Running `make test` fails with:

```
make test
...
Running "vows:all" (vows) task

  ♢ predicates

  checkLevel an item without a level settings with a critical reportLevel
    ✓ should not send
  checkLevel an item with an unknown level settings with an error reportLevel
    ✓ should not send
  checkLevel an item with an unknown level settings with an unknown reportLevel
    ✓ should send
  checkLevel an item with an unknown level settings without a reportLevel
    ✓ should send
  checkLevel an item with a warning level settings with an error reportLevel
    ✓ should not send
  checkLevel an item with a warning level settings with an info reportLevel
    ✓ should send
  checkLevel an item with a warning level settings with a warning reportLevel
    ✓ should send
/path/rollbar/node_modules/vows/lib/vows/reporters/spec.js:41
    util.print(str);
         ^

TypeError: util.print is not a function
    at Object.print (/path/rollbar/node_modules/vows/lib/vows/reporters/spec.js:41:10)
    at Object.reporter.report (/path/rollbar/node_modules/vows/bin/vows:251:31)
    at Suite.report (/path/rollbar/node_modules/vows/lib/vows/suite.js:257:37)
    at run (/path/rollbar/node_modules/vows/lib/vows/suite.js:299:22)
    at EventEmitter.<anonymous> (/path/rollbar/node_modules/vows/lib/vows/suite.js:294:25)
    at EventEmitter.emit (events.js:210:5)
    at EventEmitter.options.Emitter.emit (/path/rollbar/node_modules/vows/lib/vows.js:237:24)
    at finish (/path/rollbar/node_modules/vows/lib/vows/suite.js:384:27)
    at maybeFinish (/path/rollbar/node_modules/vows/lib/vows/suite.js:377:17)
    at Object.tryEnd (/path/rollbar/node_modules/vows/lib/vows/suite.js:351:13)
```

### linter
Seeing the linter in the `package.json`, I ran `npm run lint` but this only printed out the command line arguments for eslint (ie, not picking up the config).  If I point to the config and run on `src`, I see:

```
npm run lint -c .eslintrc src
...
/path/rollbar/.eslintrc
[...]

PR #802, opened 2019-11-22
scrubField: Add * as a 'scrub everything' value

kevinrobinson:
Hello!  Earlier today I ran into a field being reported to rollbar that I didn't expect.  I'd prefer to be able to scrub all fields, as you can with the Ruby gem.  So this PR updates `scrubFields` to allow passing the special value `"*"` that scrubs everything.

I tried to follow the guidance in https://github.com/rollbar/rollbar.js#developing, and then to run the linter and tests so I could add some test cases for this, but ran into failures running those on master.  I opened that separately in https://github.com/rollbar/rollbar.js/issues/801, so for this PR I added a test case in `utility.test.js` but couldn't run it.  I didn't add tests to the various transform test files.

### DIFF
--- a/src/utility.js
+++ b/src/utility.js
@@ -562,10 +562,12 @@ function set(obj, path, value) {
   }
 }
 
+var SCRUB_FIELDS_SCRUB_ALL_VALUE = '*';
+var SCRUB_ALL_REGEX = new RegExp('(.*)', 'i');
 function scrub(data, scrubFields) {
   scrubFields = scrubFields || [];
-  var paramRes = _getScrubFieldRegexs(scrubFields);
-  var queryRes = _getScrubQueryParamRegexs(scrubFields);
+  var paramRes = _getRegexForScrubFields(scrubFields, _makeScrubFieldRegex);
+  var queryRes = _getRegexForScrubFields(scrubFields, _makeScrubQueryParamRegex);
 
   function redactQueryParam(dummy0, paramPart) {
     return paramPart + redact();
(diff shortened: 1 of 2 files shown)


waltjones:
The Rollbar.js scrubber is different from the Ruby gem in some relevant ways. Rollbar.js gives access to scrub a lot more of the payload. This PR, when the wildcard value is used, scrubs fields that must stay present.

Perhaps the scrubber should only look at specific parts of the payload, which is how the gem works. That change would be needed before this could work.

kevinrobinson:
@waltjones Thanks!  Yeah re:
> This PR, when the wildcard value is used, scrubs fields that must stay present.

That's my intention, to expose a config that allows that.  I'm not sure what you mean by "fields that must stay present."  Are you saying this would scrub Rollbar-internal fields?  Or just that this PR doesn't support the same flexibility as the Ruby gem (eg, `scrub_whitelist`)?

I tried to do the minimal change here, but if you want to talk through broader API changes to the config first, I'm happy to do that too.  I can also defer to whatever particular API you like, since I'm sure you have more Rollbar context than I do :)

I'd be happy with any If there was a way to tell the config "scrub everything" or "scrub everything except for these specific param names" (a la `scrubWhitelist`).

Thanks for your help! 👍

waltjones:
In rollbar.js, the scrubber is applied to the entire payload:
https://github.com/rollbar/rollbar.js/blob/master/src/browser/transforms.js#L295

There are a few things that get added after that transform, but essentially that is the fully formed payload almost ready to send. So, yes this will remove everything: exception, call stack, message body, environment string, etc. It will look at every key in the payload.

The scrubber could be limited to only look at parts of the payload, similar to what the gem does. That would allow this to work, and it seems would also be more performant.

The important work when doing this is to identify which payload keys should or should not have the scrubber applied and to make it specific enough that the wildcard scrubber doesn't break things.

Don't feel like you have to take this on though. I'm happy to also look at it sometime soon.

kevinrobinson:
@waltjones thanks!  I hear what you're saying that the core of this work would be understanding and perhaps documenting or enforcing in code the various fields the transforms add in.

I'm not sure I have great context to help much then on reworking the internals or the API altogether, but if seems like this is the main flow:

```
function addTransformsToNotifier(notifier, gWindow) {
  notifier
    .addTransform(transforms.handleDomException)
    .addTransform(transforms.handleItemWithError)
    .addTransform(transforms.ensureItemHasSomethingToSay)
    .addTransform(transforms.addBaseInfo)
    .addTransform(transforms.addRequestInfo(gWindow))
    .addTransform(transforms.addClientInfo(gWindow))
    .addTransform(transforms.addPluginInfo(gWindow))
    .addTransform(transforms.addBody)
    .addTransform(sharedTransforms.addMessageWithError)
    .addTransform(sharedTransforms.addTelemetryData)
    .addTransform(sharedTransforms.addConfigToPayload)
    .addTransform(transforms.scrubPayload)
    .addTransform(sharedTransforms.userTransform(logger))
    .addTransform(sharedTransforms.addConfiguredOptions)
    .addTransform(sharedTransforms.addDiagnosticKeys)
    .addTransform(sharedTransforms.itemToPayload);
}
```

And that the code and tests for these sort of imply the overall shape of the payload.  I think it would take some reading through those, and maybe adding tests cases that explicitly test the shape of the payload, and the way the transforms interact or don't (eg, `addMessageWithError` and `addBody and `addTelemetryData` all play in the same `data.body` space, and there are implicit dependencies between them).

Let me know what you think makes sense as a next step, I'm happy to help brainstorm or dig in but changing the order of these, or changing the way scrubbing is applied to these various bits seems like it would require some testing to make sure it doesn't break folks.

waltjones:
@kevinrobinson I've opened https://github.com/rollbar/rollbar.js/issues/813 to track this, and am closing this PR.

PR #804, opened 2019-11-22
Update vows package to fix util dependency

waltjones:
Fixes: https://github.com/rollbar/rollbar.js/issues/801

Vows was previously versioned at 0.7.0, which has a call to `util.print` that results in a "not a function" error... but only sometimes. The util package in my node_modules doesn't export a `print()` function, yet everything works in my environment, and has worked in CI (which should be a clean npm environment.) Despite this seeming to work for the lucky, it is clearly wrong by code review. (I can't find an old enough version of util that it still has `print()` anywhere.)

Later versions of vows no longer use `util.print`, so this PR updates to vows 0.8.3, which looks like the last stable version.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -14,6 +14,7 @@
   },
   "plugins": [
   ],
+  "globals": { "Promise": true },
   "rules": {
     "strict": 0,
     "camelcase": [2, {"properties": "never"}],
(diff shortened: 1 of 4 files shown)


kevinrobinson:
@waltjones thanks! 😄   Yeah, this fixes vows for me (but to be clear doesn't fully resolve #801).

Issue #615, opened 2018-05-18
Addressing the package size


luke-j:
By my measurement, this package currently weighs in at around 17KB gzipped. That is quite a decent sized package by any account:
<img width="248" alt="screen shot 2018-05-18 at 10 49 50 am" src="https://user-images.githubusercontent.com/6657011/40210654-7f67dc8c-5a89-11e8-831f-918fc54c188e.png">

Are there any plans to look at slimming the size of this bundle down?

rokob:
I can give things a once over to see about getting the size down a little bit, but it will be in the 12-17 KB range given the feature set and browser support. Pull requests would be welcome if there is some cross cutting concern that would get the size down that I am missing.

luke-j:
>it will be in the 12-17 KB range given the feature set and browser support.

Perhaps consider a slim build or ES modules build for Webpack tree shaking? Also, it might be worth a separate build for more modern browsers that a consumer of this package could elect to import in place of the full package. For example, I'm only dealing with a couple of the functions with the `last 2 versions` of browser support.

dangodev:
The pre/post-gzipped weight is the real measurement, and it’s `64 KB`. That’s almost as much as React! Running a thin React application, Rollbar nearly doubles my bundle size. There’s only so many JS packages that can be “critical” before some get loaded async, or not at all, and with Rollbar kinda being “critical” it’s a shame that it impacts the user experience negatively so much.

newtack:
https://bundlephobia.com/result?p=rollbar@2.5.4

18.8 KB

If libraries such as material-ui can decrease their footprint by 15% (see here https://bundlephobia.com/result?p=@material-ui/core@4.0.0-alpha.4) while adding features, then I am sure that the great Rollbar engineers can do so too ...  :-)

lightheaded:
I agree and this is not really tolerable for high-performance applications and we're currently looking to switch the provider in case this issue is a dead end. Is there any plans to address this issue in the future? Perhaps create a lite version for critical features?

HMilbradt:
Is this still being looked into?  This is the largest package in one of our projects, and its size is contributing to a negative impact on pagespeed scores.

waltjones:
I've assigned to myself. I don't have an ETA or a good idea about how much is possible in the short term but I'll look at it. Feel free to send any ideas or feedback.

waltjones:
[...]

PR #817, opened 2020-01-18
Option to reduce browser bundle size

waltjones:
fixes: https://github.com/rollbar/rollbar.js/issues/615

Allows the current standard package for browser to continue to work as is. Provides the ability for Webpack and other builders to use `require(‘rollbar/src/browser/core’)` instead. Just using this will build a minimal Rollbar.js with telemetry and other optional features not included. Each of these features can be included a la carte to provide the package needed at the smallest possible size.

Minimal example:
```
var Rollbar = require('rollbar/src/browser/core');
rollbar = Rollbar.init(config);
rollbar.log(‘Hello’);
```

Example with all optional features included:
```
var Rollbar = require('rollbar/src/browser/core');
var telemeter = require('rollbar/src/telemetry');
var instrumenter = require('rollbar/src/browser/telemetry');
var polyfillJSON = require('rollbar/vendor/JSON-js/json3');
var wrapGlobals = require('rollbar/src/browser/wrapGlobals');
var scrub = require('rollbar/src/scrub');
var truncation = require('rollbar/src/truncation');

Rollbar.setComponents({
  telemeter: telemeter,
  instrumenter: instrumenter,
  polyfillJSON: polyfillJSON,
  wrapGlobals: wrapGlobals,
  scrub: scrub,
  truncation: truncation
});
 
rollbar = Rollbar.init(config);
 
rollbar.log(‘Hello’);
```

The following size estimates are based on a vanilla webpack 4 config, and measure the pre-zip minified size.

The minified size of the minimal package (browser/core only) is 49K, down from 74K for a 34% reduction.

**Base telemeter** - The previously always-on telemetry capture for Rollbar message logging.
`require(rollbar/src/telemetry)` (saves ~3K minified)

**Browser telemetry** - All browser telemetry capabilities: Console, network, click events, etc.
`require(rollbar/src/browser/telemetry)` (saves ~14K minified)

**JSON polyfill** - Some pre-2009 browsers require a polyfill for JSON.stringify and JSON.parse.
`require(rollbar/vendor/JSON-js/json3)` (saves ~7K minified)

**Wrap globals** - This is already disabled by default. Most people would benefit from not bundling.
` require('rollbar/src/browser/wrapGlobals')` (saves ~1K)

**Scrub** - By making parts of the payload optional, there may be no remaining need for scrubbing in specific configurations. Or the remaining payload isn’t considered sensitive by the user. 
`require('rollbar/src/scrub')` (saves ~1K)

**Truncation** - Not likely needed when telemetry isn't present.  
`require('rollbar/src/truncation')` (saves ~1.6K)

**package.json** - Rollbar.js builds its /dist/ using webpack's `DefinePlugin` to load defaults from package.json. When building from /src/ though, all of package.json gets required. This PR refactors the defaults so it's not necessary to `require('package.json')`. (saves ~3K minified)

### For reviewers:
This is a large PR, but straightforward.

* Move the contents of browser/rollbar.js to browser/core.js. browser/rollbar.js is still the entry point for dist/rollbar.min.js. It now calls `Rollbar.setComponents()` with all optional components included.
* Because browser/rollbar.js uses `Rollbar.setComponents()`, all browser test automation now tests the new componentized build.
* Each component is extracted by lifting the require out of rollbar.js and into the app, then passing in as a `setComponent` option.
* Tests are added for the minimal browser/core.js build (with no optional components added.)

Each of the above is in its own commit, to make them easier to review in isolation.

### Future work
This PR does not create a minimal build target in dist (e.g. dist/rollbar.core.min.js). This can be added in the future if there's demand. The use cases we currently know about are focused around webpack builds, and similar build tools.

There's been discussion about having scoped packages (e.g. @rollbar/browser, @rollbar/core, etc.). This work is a logical precursor to that and can be used to generate those build targets.


src/api.js
```diff
@@ -30,13 +29,14 @@ var defaultOptions = {
...
-  this.url = u;
-  this.jsonBackup = j;
+  this.transport = transport;
+  this.url = urllib;
+  this.truncation = truncation;
+  this.jsonBackup = jsonBackup;
   this.accessToken = options.accessToken;
-  this.transportOptions = _getTransport(options, u);
+  this.transportOptions = _getTransport(options, urllib);
 }
```
austenLacy, line 40: 🙌

src/browser/transforms.js
```diff
@@ -309,11 +309,15 @@ function errorClass(stackInfo, guess, options) {
   }
 }
 
-function scrubPayload(item, options, callback) {
-  var scrubFields = options.scrubFields || [];
-  var scrubPaths = options.scrubPaths || [];
-  item.data = _.scrub(item.data, scrubFields, scrubPaths);
-  callback(null, item);
+function addScrubber(scrub) {
```
austenLacy, line 312: not a big deal, but maybe `scrubFn` over `scrub` just to add some clarity

FabianSellmann:
Will these modules be eventually included in the type definitions?

waltjones:
@FabianSellmann Yes. https://github.com/rollbar/rollbar.js/issues/880

PR #831, opened 2020-03-05
Add upload to s3 workflow

austenLacy:
We are moving to use AWS CloudFront backed by S3 as our CDN moving forward. This automates the upload of the various rollbar.js distributions to S3.


.github/workflows/workflow.yml
```diff
@@ -0,0 +1,21 @@
...
+    - uses: jakejarvis/s3-sync-action@v0.5.1
+      with:
+        args: --follow-symlinks --delete
+      env:
+        AWS_S3_BUCKET: ${{ secrets.AWS_S3_BUCKET }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        AWS_REGION: 'us-east-1'
+        SOURCE_DIR: 'dist'
+        DEST_DIR: rollbarjs/$GITHUB_REF # GITHUB_REF is the tag of the release
```
jondeandres: Now we use `rollbar.js/2.2.8/` (for ex) in the cloudflare URL. What's expected with this in combination with cloudfront?

I guess will be `/rollbarjs/v2.2.8` since we do tags with `v` as prefix no?

waltjones: It ends up being `/rollbarjs/refs/tags/v2.2.8`:
https://cdn.rollbar.com/rollbarjs/refs/tags/v2.14.10/rollbar.min.js

jondeandres: I see, there is no way to simplify that schema I guess?

waltjones: We're using jakejarvis/s3-sync-action as is, and it uses `$GITHUB_REF` as is. With a fork and some custom code we could do better.

I thought the same thing when I first saw it, but also noted that the cdnjs URLs had a similar issue by containing `/ajax/libs/`. https://cdnjs.cloudflare.com/ajax/libs/rollbar.js/2.14.4/rollbar.min.js

So we could do better, but are not worse off than what we had.

jondeandres: :+1: :100:

PR #832, opened 2020-03-05
Added handler for Azure Function runtime

ahuggins-nhs:
Azure functions operate similarly to AWS Lambda in that uncaughtException events are not emitted.
Support for these patterns are added as convenience methods to pass-through the expected results from both sync and async patterns.
https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-error-pages#use-structured-error-handling
https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node

### DIFF
--- a/src/server/rollbar.js
+++ b/src/server/rollbar.js
@@ -281,6 +281,41 @@ Rollbar.errorHandler = function() {
   }
 };
 
+Rollbar.prototype.azureFunctionHandler = function(handler) {
+  const AsyncFunction = (async () => {}).constructor;
+  if (handler instanceof AsyncFunction) {
+    return this.asyncAzureFunctionHandler(handler);
+  }
+  return this.syncAzureFunctionHandler(handler);
+};
+
+Rollbar.prototype.asyncAzureFunctionHandler = function(handler) {
+  const self = this;
+  return async function(context, ...args) {
+    try {
+      return await handler(context, ...args);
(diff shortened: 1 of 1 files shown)


ahuggins-nhs:
This was directly inspired and riffed from https://docs.rollbar.com/docs/aws-lambda. For Azure, this wasn't very complex as they depend on the first-class async function to pass a result back to the runtime or to return void. For sync functions, the developer calls `context.done()` to finalize the sync method. In either case, we're just passing plain object/void back.

ahuggins-nhs:
@waltjones, do you have any thoughts on this? We're looking at rolling out Rollbar more widely across out Azure applications. Currently we're modifying the prototype to add this code in one or two Azure Function App projects, but it would be nice to have this included in the upstream.

waltjones:
@ahuggins-nhs Apologies for the delay. The next release will be when I'm able to loop back to rollbar.js.

Superficially, the PR looks good. Tests would be good. (I realize there are not tests for the LambdaHandler, and that needs to be improved.)

ahuggins-nhs:
> Superficially, the PR looks good. Tests would be good. (I realize there are not tests for the LambdaHandler, and that needs to be improved.)

I'll give some thought on writing tests. Our workload has picked up considerably, but I might be able to carve some time to write them in the next week. Thanks, @waltjones !

PR #841, opened 2020-03-27
Fix error while initializing LRU cache

Pita:
Running rollbar on aws lambda with node version 12.16.1 gave me the following error:

```
Class constructor LRUCache cannot be invoked without 'new'
```

This pull request fixes it


Pita:
Travis has finished checking this, not sure why its still yellow

Pita:
Hello @waltjones, could you have a look at this. My company currently works on a patched version because else we can't use rollbar

waltjones:
@Pita Will aim to get a release out today.

BTW, the Travis build is green. Looks like Github never synced to its final state.

waltjones:
This is released in v2.15.2. 

https://github.com/rollbar/rollbar.js/releases/tag/v2.15.2

PR #843, opened 2020-04-14
Only build stack for http errors when the feature is enabled

waltjones:
This PR ensures stacks for reporting xhr and fetch errors are built only when that feature is enabled.

Calling `stack()` on an Error object causes it to show up at `Error.prepareStackTrace`. That's unavoidable, but limiting it as much as possible helps the anonymous errors feature (which relies on `Error.prepareStackTrace`) work more reliably in more situations.

### DIFF
--- a/src/browser/telemetry.js
+++ b/src/browser/telemetry.js
@@ -296,7 +296,7 @@ Instrumenter.prototype.instrumentNetwork = function() {
         } else {
           xhr.onreadystatechange = onreadystatechangeHandler;
         }
-        if (xhr.__rollbar_xhr) {
+        if (xhr.__rollbar_xhr && self.trackHttpErrors()) {
           xhr.__rollbar_xhr.stack = (new Error()).stack;
         }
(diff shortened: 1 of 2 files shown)


src/browser/telemetry.js
```diff
@@ -354,7 +354,9 @@ Instrumenter.prototype.instrumentNetwork = function() {
           }
         }
         self.captureNetwork(metadata, 'fetch', undefined);
-        metadata.stack = (new Error()).stack;
+        if (self.trackHttpErrors()) {
+          metadata.stack = (new Error()).stack;
```
jondeandres, line 358: Should we send `stack = null` here or is ok not sending the `stack` property at all?

waltjones: I think not setting at all would be the preference.

jondeandres: :+1:

waltjones:
@jondeandres Thank you!

PR #847, opened 2020-05-06
Enable onSendCallback and transform callabck for react-native

waltjones:
References https://github.com/rollbar/rollbar-react-native/issues/109

This PR enables `onSendCallback` and `transform` for the react-native target when set in the configuration.

This PR also updates the react-native target to used the shared implementation of the predicates (checkIgnore, checkLevel) which include the support for `onSendCallback`. The react-native specific implementation for these was added before the shared predicates existed. When shared predicates were added, the react-native implementation didn't get removed.

2nd commit addresses a change in behavior in Travis' headless chrome.


vselvarajijay:
This looks good to me

Issue #837, opened 2020-03-12
Incorrect `LambdaHandler` type


AndKiel:
Exported `LambdaHandler` type allows only sync functions even though AWS also supports async lambdas:

https://github.com/rollbar/rollbar.js/blob/bf9a2efcdb42dd5d1f8e2fe4883011cf286fc2e5/index.d.ts#L35

Rollbar's internal logic also seems to support both sync and async:

https://github.com/rollbar/rollbar.js/blob/63c2f64d1fa9b99525fa62d2fa9110769f85e20e/src/server/rollbar.js#L284

but due to the incorrect type declaration linting and/or building of a project can fail.

Possibly Rollbar could use https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/handler.d.ts types for consistency with aws-lambda package.

PR #849, opened 2020-05-07
Fix type definition to work with async handler functions

waltjones:
Fixes: https://github.com/rollbar/rollbar.js/issues/837

Uses the definition from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/handler.d.ts as suggested in the linked issue.


mrunalk:
LGTM.

PR #850, opened 2020-05-08
Diagnostic improvements

waltjones:
Add a few things to the `notifier.diagnostic` key in the payload.
* is_uncaught
* Callback functions from config, if present (`transform`, `onSendCallback`, `checkIgnore`)
* Caller's type signature for `Rollbar.log()`

Payload keys are snake_case, except option keys within `configured_options` which retain their original javascript case.

### DIFF
--- a/src/transforms.js
+++ b/src/transforms.js
@@ -75,9 +75,22 @@ function addConfigToPayload(item, options, callback) {
   callback(null, item);
 }
 
+function addFunctionOption(options, name) {
+  if(_.isFunction(options[name])) {
+    options[name] = options[name].toString();
+  }
+}
+
 function addConfiguredOptions(item, options, callback) {
-  delete options._configuredOptions.accessToken;
-  item.data.notifier.configured_options = options._configuredOptions;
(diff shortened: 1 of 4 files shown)


src/transforms.js
```diff
@@ -88,6 +101,11 @@ function addDiagnosticKeys(item, options, callback) {
     diagnostic.is_anonymous = true;
   }
 
+  if (item._isUncaught) {
+    diagnostic.is_uncaught = item._isUncaught;
```
austenLacy, line 105: does this need to be a boolean?

if so we should probably "shebang" it with
```
diagnostic.is_uncaught = !!item._isUncaught;
```

austenLacy:
@waltjones do we need to update any of the Typescript types for these changes?

src/transforms.js
waltjones: This key is diagnostic only and is never evaluated, so shouldn't be an issue.

waltjones:
Currently we don't have TypeScript definitions for the payload data structure.

PR #851, opened 2020-05-08
Allow configuration of stack trace limit

waltjones:
The default stack trace length in Chrome and Node is 10 frames, which often doesn't include the frames related to the cause of the error.

Chrome, Safari, Node, and some other environments allow increasing the number of stack frames by setting `Error.stackTraceLimit`.

The config option, off by default, can be set at `options.stackTraceLimit`. Changing the default would impact grouping, since stack traces for the same error may now be a different length. Therefore the current behavior is preserved as the default.

### DIFF
--- a/test/browser.rollbar.test.js
+++ b/test/browser.rollbar.test.js
@@ -477,7 +477,7 @@ describe('options.captureUncaught', function() {
     done();
   });
 
-    it('should capture exta frames when stackTraceLimit is set', function(done) {
+  it('should capture exta frames when stackTraceLimit is set', function(done) {
     var server = window.server;
     stubResponse(server);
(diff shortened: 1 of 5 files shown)


test/browser.rollbar.test.js
```diff
@@ -477,6 +477,39 @@ describe('options.captureUncaught', function() {
     done();
   });
 
+    it('should capture exta frames when stackTraceLimit is set', function(done) {
```
austenLacy, line 480: indentation is off a little bit here, but small potatoes ;)

waltjones: Oops. I'll fix it.

Issue #853, opened 2020-05-13
TypeScript compilation error: Cannot find name 'Context'


alecgibson:
We seem to have issues bumping rollbar.js from v2.15.0 to v2.16.0. Our TypeScript compilation fails with:

```
node_modules/rollbar/index.d.ts:37:18 - error TS2304: Cannot find name 'Context'.

37         context: Context,
                    ~~~~~~~

node_modules/rollbar/index.d.ts:38:19 - error TS2315: Type 'Callback' is not generic.

38         callback: Callback<TResult>,
```

In case it's useful, here are our dependencies:

```json
{
  "dependencies": {
    "chalk": "^4.0.0",
    "rollbar": "^2.16.0"
  },
  "devDependencies": {
    "@types/chai": "^4.2.11",
    "@types/mocha": "^7.0.2",
    "@types/node": "^13.11.1",
    "@types/proxyquire": "^1.3.28",
    "@types/sinon": "^9.0.0",
    "@types/sinon-chai": "^3.2.4",
    "chai": "^4.2.0",
    "eslint": "^6.8.0",
    "mocha": "^7.1.1",
    "proxyquire": "^2.1.3",
    "sinon": "^9.0.2",
    "sinon-chai": "^3.5.0",
    "ts-node": "^8.8.2",
    "typescript": "^3.8.3"
  }
}
```

alecgibson:
This is the suspect commit: https://github.com/rollbar/rollbar.js/commit/17810b0d39cba2af08085460647bb0fd8ab953d5

It's only from 6 days ago, and adds the lines that are throwing.

jiangzhuo:
This is LambdaHandler I think, maybe `Context`  is from `aws-lambda` 
@alecgibson I think your solution `TContext` is better.

AndKiel:
Yeah, `Context` is `aws-lambda` thing and it's quite specific:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/handler.d.ts#L94

PR #854, opened 2020-05-13
Fix TypeScript compilation errors

alecgibson:
Fixes: https://github.com/rollbar/rollbar.js/issues/853

At the moment, the `LambdaHandler` type does not correctly compile,
because:

 - the `Context` type is not defined
 - the `Callback` type is not a generic, but is treated as such

This change fixes these errors by:

 - adding an optional `TContext` to the `LambdaHandler` generic types
 - adding an optional generic typing to the `Callback` type

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -32,9 +32,9 @@ declare class Rollbar {
 }
 
 declare namespace Rollbar {
-    export type LambdaHandler<TEvent = any, TResult = any> = (
+    export type LambdaHandler<TEvent = any, TResult = any, TContext = any> = (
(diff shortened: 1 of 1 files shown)


alecgibson:
NB: I don't know if `Context` was already meant to have been defined somewhere? Where is this thing meant to have come from?

waltjones:
Will publish a release today.

waltjones:
@alecgibson Thank you for the PR.

v2.16.1 is published. https://github.com/rollbar/rollbar.js/releases/tag/v2.16.1

PR #856, opened 2020-05-15
Update Readme with correct Travis build status link

mrunalk:
Using the latest link as described here, https://docs.travis-ci.com/user/status-images/#travis-ci-pages-show-the-default-branchs-result


mrunalk:
Good catch. Thanks for pointing out.

PR #860, opened 2020-05-28
Add opentracing integration

austenLacy:
These changes allow for a rollbar.js user to inject an OpenTracing tracer object as a configuration option. 

Whenever an error or message is logged via Rollbar we will check if the client has as tracer set. If that tracer is valid then we set the new occurrence UUID as metadata on the opentracing span as well as add the opentracing trace ID and span ID to the occurrence as custom metadata.

### DIFF
--- a/src/browser/rollbar.js
+++ b/src/browser/rollbar.js
@@ -18,7 +18,11 @@ function Rollbar(options, client) {
   this.options = _.handleOptions(defaultOptions, options);
   this.options._configuredOptions = options;
   var api = new API(this.options, transport, urllib);
-  this.client = client || new Client(this.options, api, logger, 'browser');
+
+  // remove tracer object from options so we don't try to parse it
+  var tracer = options.tracer || null;
+  delete options.tracer;
+  this.client = client || new Client(this.options, api, logger, 'browser', tracer);
 
   var gWindow = _gWindow();
   var gDocument = (typeof document != 'undefined') && document;
--- a/src/react-native/rollbar.js
+++ b/src/react-native/rollbar.js
@@ -23,7 +23,12 @@ function Rollbar(options, client) {
   delete this.options.maxItems;
   this.options.environment = this.options.environment || 'unspecified';
   var api = new API(this.options, transport, urllib);
-  this.client = client || new Client(this.options, api, logger, 'react-native');
+
+  // remove tracer object from options so we don't try to parse it
+  var tracer = options.tracer || null;
+  delete options.tracer;
+  this.client = client || new Client(this.options, api, logger, 'react-native', tracer);
+
   addTransformsToNotifier(this.client.notifier);
   addPredicatesToQueue(this.client.queue);
 }
--- a/test/browser.rollbar.test.js
+++ b/test/browser.rollbar.test.js
@@ -5,6 +5,31 @@
 
 var Rollbar = require('../src/browser/rollbar');
 
+const DUMMY_TRACE_ID = 'some-trace-id';
+const DUMMY_SPAN_ID = 'some-span-id';
+
+const ValidOpenTracingTracerStub = {
+  scope: () => {
+    return {
+      active: () => {
+        return {
+          setTag: () => { },
+          context: () => {
+            return {
+              toTraceId: () => DUMMY_TRACE_ID,
+              toSpanId: () => DUMMY_SPAN_ID
+            }
+          }
+        }
+      }
+    }
+  }
+};
+
+const InvalidOpenTracingTracerStub = {
+  foo: () => { }
+};
+
 function TestClientGen() {
   var TestClient = function() {
     this.transforms = [];
@@ -35,6 +60,7 @@ function TestClientGen() {
       this.options = o;
       this.payloadData = payloadData;
     };
+    this.tracer = ValidOpenTracingTracerStub;
   };
 
   return TestClient;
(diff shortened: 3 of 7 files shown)


src/react-native/rollbar.js
```diff
@@ -23,7 +23,12 @@ function Rollbar(options, client) {
   delete this.options.maxItems;
   this.options.environment = this.options.environment || 'unspecified';
   var api = new API(this.options, transport, urllib);
-  this.client = client || new Client(this.options, api, logger, 'react-native');
+
+  // remove tracer object from options so we don't try to parse it
+  var tracer = options.tracer || null;
+  delete options.tracer;
+  this.client = client || new Client(this.options, api, logger, 'react-native', tracer);
```
vselvarajijay, line 30: Logic looks similar to what's in src/browser/rollbar.js. Maybe worth refactoring into a shared function to set this.client

test/browser.rollbar.test.js
```diff
@@ -153,6 +179,24 @@ describe('Rollbar()', function() {
 
     done();
   });
+
+  it('should have a tracer if valid tracer is provided', function(done) {
```
vselvarajijay, line 183: Should we also have a test when no tracer object is provided?

test/react-native.rollbar.test.js
```diff
@@ -190,6 +214,24 @@ describe('Rollbar()', function() {
 
     done();
   });
+
+  it('should have a tracer if valid tracer is provided', function (done) {
```
vselvarajijay, line 218: Should we also have a test when no tracer object is provided?

test/browser.rollbar.test.js
austenLacy: @vselvarajijay that's the same as the invalid tracer object IMO

src/react-native/rollbar.js
austenLacy: Each implementation (node, react-native, browser) don't inherit from some shared class so their instantiation of the client is unique to their implementation so I don't think a shared function would make sense since the client has to live as an attribute to the individual implementation.

test/server.rollbar.test.js
```diff
@@ -116,6 +142,24 @@ vows.describe('rollbar')
         'should set configured options': function(r) {
           assert.equal('fake-env', r.options._configuredOptions.environment);
         }
+      },
+      'with valid tracer': {
```
vselvarajijay, line 146: again not sure if we need a 'without tracer' test

test/server.rollbar.test.js
```diff
@@ -129,12 +173,65 @@ vows.describe('rollbar')
           assert.equal('new-env', r.options._configuredOptions.environment);
           assert.equal(false, r.options._configuredOptions.captureUncaught);
         }
+      },
+      'with valid tracer': {
```
vselvarajijay, line 177: again not sure if we need a 'without tracer' test

test/server.rollbar.test.js
```diff
@@ -129,12 +173,65 @@ vows.describe('rollbar')
...
+          rollbar.configure({ tracer: InvalidOpenTracingTracerStub });
+          return rollbar;
+        },
+        'should configure tracer': function(r) {
+          assert.equal(null, r.client.tracer);
+        }
       }
     },
+    'addTracingInfo': {
+      'with valid tracer': {
```
vselvarajijay, line 199: again not sure if we need a 'without tracer' test

test/server.rollbar.test.js
```diff
@@ -129,12 +173,65 @@ vows.describe('rollbar')
...
+        'should add trace ID and span ID to custom field on item even if already has some custom fields': function(r) {
+          const item = { uuid: 'some-uuid', custom: { foo: 'foo' } };
+          r.client._addTracingInfo(item);
+          assert.equal(DUMMY_TRACE_ID, item.custom.opentracing_trace_id);
+          assert.equal(DUMMY_SPAN_ID, item.custom.opentracing_span_id);
+          assert.equal('foo', item.custom.foo);
+        }
+      },
+      'with invalid tracer': {
+        topic: function() {
```
vselvarajijay, line 219: again not sure if we need a 'without tracer' test

test/server.rollbar.test.js
waltjones: I agree with @austenLacy . This is covered by the invalid tracer test.

src/browser/rollbar.js
```diff
@@ -18,7 +18,11 @@ function Rollbar(options, client) {
   this.options = _.handleOptions(defaultOptions, options);
   this.options._configuredOptions = options;
   var api = new API(this.options, transport, urllib);
-  this.client = client || new Client(this.options, api, logger, 'browser');
+
+  // remove tracer object from options so we don't try to parse it
+  var tracer = options.tracer || null;
+  delete options.tracer;
```
waltjones, line 24: If this is to prevent serialization to `configured_options`, consider putting a placeholder that is serializable (even just a message string) to indicate whether a valid tracer was found. This will help support in the future if they're looking into a tracer related occurrence.

src/react-native/rollbar.js
```diff
@@ -23,7 +23,12 @@ function Rollbar(options, client) {
   delete this.options.maxItems;
   this.options.environment = this.options.environment || 'unspecified';
   var api = new API(this.options, transport, urllib);
-  this.client = client || new Client(this.options, api, logger, 'react-native');
+
+  // remove tracer object from options so we don't try to parse it
+  var tracer = options.tracer || null;
+  delete options.tracer;
```
waltjones, line 29: Each of browser, server, and react-native does this the same way. This could be moved into the `Client` constructor, keeping it in one place, and then the additional argument to the constructor isn't needed either.

waltjones:
Changes look good. :+1:

src/server/transforms.js
```diff
@@ -13,7 +13,8 @@ function baseData(item, options, callback) {
     language: 'javascript',
     framework: item.framework || options.framework,
     uuid: item.uuid,
-    notifier: JSON.parse(JSON.stringify(options.notifier))
+    notifier: JSON.parse(JSON.stringify(options.notifier)),
+    custom: item.custom
```
renatochaves, line 17: This results in duplicated properties in Occurrences view... Could we revert this part?

waltjones: I think the intent is to support the `custom` key as defined by the API. https://explorer.docs.rollbar.com/#operation/create-item

Maybe a config option for this is the best way to meet both needs.

PR #869, opened 2020-06-29
Allow scrubbing by full path in payload

waltjones:
This PR adds `config.scrubPaths`, which allows scrubbing specific key locations by specifying the full path of the key. (When using `config.scrubFields`, all matching key names in the payload will be scrubbed regardless of location.)

Example:
```
config.scrubPaths = [
  'request.query_string',
  'client.javascript.plugins'
]
```

This makes it possible to scrub 'request.body', whereas adding 'body' to `config.scrubFields` would scrub every `body` key in the payload.

By adding `scrubPaths` as a separate key from `scrubFields`, fields that contain dots will still work and paths that don't (root keys) will also work.

This PR does not support paths that contain arrays. That can be added later if there is sufficient interest.

### DIFF
--- a/src/utility.js
+++ b/src/utility.js
@@ -602,8 +602,15 @@ function set(obj, path, value) {
   }
 }
 
-function scrub(data, scrubFields) {
+function scrub(data, scrubFields, scrubPaths) {
   scrubFields = scrubFields || [];
+
+  if (scrubPaths) {
+    for (var i = 0; i < scrubPaths.length; ++i) {
+      scrubPath(data, scrubPaths[i]);
+    }
+  }
+
   var paramRes = _getScrubFieldRegexs(scrubFields);
   var queryRes = _getScrubQueryParamRegexs(scrubFields);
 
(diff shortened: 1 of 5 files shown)


src/utility.js
```diff
@@ -647,6 +654,22 @@ function scrub(data, scrubFields) {
...
+  var keys = path.split('.');
+  var last = keys.length - 1;
+  try {
+    for (var i = 0; i <= last; ++i) {
+      if (i < last) {
+        obj = obj[keys[i]];
+      } else {
+        obj[keys[i]] = redact();
+      }
+    }
```
austenLacy, line 667: took me a minute to figure out exactly how this works, but it's clever

test/utility.test.js
```diff
@@ -702,6 +702,33 @@ describe('scrub', function() {
...
+      'nowhere', // path not found
+      'a.b.foo', // nested path
+      'a.password', // nested path
+      'secret' // root path
+    ];
+
+    var result = _.scrub(data, [], scrubPaths);
+
+    expect(result.a.b.bar).to.eql('stuff');
+    expect(result.a.b.foo).to.not.eql('secret');
```
austenLacy, line 727: a nitpick, but maybe a better check of the actual redaction would be to see if it's
```
expect(result.a.b.foo).to.eql('*******');
```

waltjones: Good call. I had reused the same pattern from other tests, assuming the `redact()` function isn't available here. Actually it is, so I've updated all existing tests as well.

PR #870, opened 2020-07-02
Send error context as custom data

waltjones:
Sometimes relevant debugging context is known at the time the error is thrown rather than at the time it is caught or reported. This PR enables detecting context data on the error object and merging it with the custom data in the Rollbar occurrence.

To use the feature:
* Set `options.addErrorContext = true`.
* Set `error.rollbarContext` with the data to be included in the custom data for the occurrence.

For trace chains, context data will be merged from any errors where it is present.

### DIFF
--- a/src/utility.js
+++ b/src/utility.js
@@ -506,6 +506,27 @@ function setCustomItemKeys(item, custom) {
   }
 }
 
+function addErrorContext(item, errors) {
+  var custom = item.data.custom || {};
+  var contextAdded = false;
+
+  try {
+    for (var i = 0; i < errors.length; ++i) {
+      if (errors[i].hasOwnProperty('rollbarContext')) {
+        custom = merge(custom, errors[i].rollbarContext);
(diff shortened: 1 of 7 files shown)


src/utility.js
```diff
@@ -506,6 +506,27 @@ function setCustomItemKeys(item, custom) {
   }
 }
 
+function addErrorContext(item, errors) {
+  var custom = item.data.custom || {};
+  var contextAdded = false;
+
+  try {
+    for (var i = 0; i < errors.length; ++i) {
+      if (errors[i].hasOwnProperty('rollbarContext')) {
```
austenLacy, line 515: is it safe to assume `errors[i]` is always an object? otherwise we might want a `typeof` check here before the `hasOwnProperty` just in case

waltjones: The code that calls this is already assuming a valid Error object, so should be safe to continue that assumption here since it isn't an external interface.

Issue #873, opened 2020-07-14
Missing "addErrorContext" to configuration TS type


Mister-Corn:
PR #870 adds a new option, `addErrorContext`, which can be added to the configuration object when initializing Rollbar. This new option is not listed within the`Rollbar.Configuration` type in `index.d.ts`. 

Should that be added in? (_If so, may I have the honor?_ 😉 )

waltjones:
Yes! :+1:

PR #874, opened 2020-07-15
Add "addErrorContext" in config type

Mister-Corn:
PR #870 adds a new option, addErrorContext, which can be added to the configuration object when initializing Rollbar. This option will be included in the `Rollbar.Configuration` type in `index.d.ts`, to ensure people who type their configuration object to be able to use this option.

Fixes: #873


waltjones:
@Mister-Corn thank you!

Issue #881, opened 2020-08-04
checkIgnore "isUncaught" argument is always `false`


highspotjeremy:
The `isUncaught` (first) argument in the `checkIgnore` function call is always `false`. This change appears to be the result of PR #850. Specifically, [this line](https://github.com/rollbar/rollbar.js/pull/850/files#diff-c854b4a38519d3019f2e50f74e73bf5fR106) deletes the `_isUncaught` property on the item before the `userCheckIgnore` predicate is being called. Prior to this change, that property was deleted in that predicate method.

Code sample:
```javascript
const options = {
  checkIgnore: (isUncaught) => {
    console.log(isUncaught);
  }
};
Rollbar.configure(options);
window.onerror(new Error('test'));
```

This should produce a `true` in the console since the `window.onerror` hook is getting called, but it instead produces `false`. (I can replicate this by doing an actual `throw`.)

waltjones:
@highspotjeremy Thank you.

waltjones:
@highspotjeremy Fix released in v2.19.2.

highspotjeremy:
@waltjones - thank you!

PR #882, opened 2020-08-06
Ensure checkIgnore receives isUncaught

waltjones:
Fixes: https://github.com/rollbar/rollbar.js/issues/881

Don't delete `item._isUncaught` until checkIgnore predicate.

Also adds test that runs all transforms and predicates (Other checkIgnore tests only run predicates), and resets the `captureUncaught` state between tests.


waltjones:
@ajtran Yes.

PR #883, opened 2020-08-07
Make setRequestHeader safe after async open

waltjones:
The spec for `XMLHttpRequest` requires `open()` to be called before `setRequestHeader()`, however `open()` may be called async and in that case Rollbar's handler for `setRequestHeader()` may execute first.

This PR refactors the initialization for both `open()` and `setRequestHeader()` to make async open safe in all cases.

**Testing**
The existing tests already use async open, but the actual execution order is browser dependent. This error and the fix have been tested manually with a browser (Chrome 73) and app that exercise the condition.

### DIFF
--- a/src/browser/telemetry.js
+++ b/src/browser/telemetry.js
@@ -169,15 +169,20 @@ Instrumenter.prototype.instrumentNetwork = function() {
     replace(xhrp, 'open', function(orig) {
       return function(method, url) {
         if (_.isType(url, 'string')) {
-          this.__rollbar_xhr = {
-            method: method,
-            url: url,
-            status_code: null,
-            start_time_ms: _.now(),
-            end_time_ms: null
-          };
-          if (self.autoInstrument.networkRequestHeaders) {
-            this.__rollbar_xhr.request_headers = {};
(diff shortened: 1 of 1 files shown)


src/browser/telemetry.js
```diff
@@ -186,12 +191,21 @@ Instrumenter.prototype.instrumentNetwork = function() {
...
+          if (self.autoInstrument.networkRequestHeaders) {
+            if (!this.__rollbar_xhr.request_headers) {
+              this.__rollbar_xhr.request_headers = {};
+            }
+            this.__rollbar_xhr.request_headers[header] = value;
+          }
+          // We want the content type even if request header telemetry is off.
+          if (header.toLowerCase() === 'content-type') {
+            this.__rollbar_xhr.request_content_type = value;
+          }
```
ajtran, line 208: Looks like this executes only when `(_.isType(header, 'string') && _.isType(value, 'string'))` which makes sense to me, but i believe is different than the older code. Just want to make sure that this is intended?

waltjones: Yes :+1:

Issue #885, opened 2020-08-15
Crash on react-native 0.63.2 on hermes engine enabled android release build


ahmaddehnavi:
rollbar version : **2.19.2**
rollbar-react-native version: **0.9.0**
react-native version **0.63.2 (upgraded from 0.61.5)**
hermes-engine version: **0.5.1** 

variable values:
```
console.log('polyfillJSON',polyfillJSON) // 'polyfillJSON', undefined
console.log('JSON',JSON)  //  'JSON', {}
console.log('JSON.stringify',JSON.stringify) //  'JSON.stringify', [Function: stringify]
console.log('JSON.parse',JSON.parse) //  'JSON.parse', [Function: parse]
```

the app crashes on `setupJSON` method **stack traces.**
```
    com.facebook.react.common.JavascriptException: TypeError: undefined is not a function, js engine: hermes, stack:
    setupJSON@1:957545
    Rollbar@1:951045
    Client@1:947927
    AppLogger@1:945314
    anonymous@1:944357
    loadModuleImplementation@1:62729
    guardedLoadModule@1:62389
    metroRequire@1:62015
    anonymous@1:924026
    loadModuleImplementation@1:62729
    guardedLoadModule@1:62389
    metroRequire@1:62015
    anonymous@1:916131
    loadModuleImplementation@1:62729
    guardedLoadModule@1:62389
    metroRequire@1:62015
    anonymous@1:744717
    loadModuleImplementation@1:62729
    guardedLoadModule@1:62389
    metroRequire@1:62015
    anonymous@1:613562
    loadModuleImplementation@1:62729
    guardedLoadModule@1:62389
    metroRequire@1:62015
    anonymous@1:606930
    loadModuleImplementation@1:62729
    guardedLoadModule@1:62389
    metroRequire@1:62015
    anonymous@1:571354
    loadModuleImplementation@1:62729
    guardedLoadModule@1:62389
    metroRequire@1:62015
    anonymous@1:69136
    loadModuleImplementation@1:62729
    guardedLoadModule@1:62346
    metroRequire@1:62015
    global@1:61641
```

and the error has gone when I commented ` polyfillJSON(RollbarJSON);` line so it can be fixed by check polyfillJSON to be defined

waltjones:
Fix released in v2.19.3.

PR #886, opened 2020-08-15
fix crash on react-native 63.2 when hermes engine enabled

ahmaddehnavi:
fix https://github.com/rollbar/rollbar.js/issues/885


ahmaddehnavi:
fixed on v2.19.3

PR #902, opened 2020-11-04
Enable node.js stack locals capture

waltjones:
## Description of the change

Enables capture of local stack variables in Node.js. To enable, set `locals: true` in the config.

This feature uses the Node Inspector API to access debugger data at runtime. The Inspector API is a wrapper for the V8 Inspector interface. While the V8 interface relies on `send()` and `receive()` functions that can't be accessed from Node, the Node Inspector API provides access to these via the `post()` method on the Inspector Session instance. Using `session.post()`, all of the DevTools protocol can be accessed, including the `Debugger` and `Runtime` interfaces, which are used here.

The initial version of this feature supports both caught and uncaught exceptions and promise rejections. Nested exceptions and trace chains are supported.

The DevTools protocol allows retrieving nested variables to an arbitrary depth. However, the initial version of this feature only implements the equivalent of depth = 0. A future PR may enable arbitrary depth with a config option to set the depth.

Unexpanded objects are identified by their class, in keeping with the example in the Rollbar API docs. (https://explorer.docs.rollbar.com/#tag/Item)

Example:
```
trace: {
  frames: [
    {
      // ...
      locals: {
        foo: 'bar', // immediate string value
        obj: '<Foo object>' // instance of Foo
      }
    }
  ]
}
```

The feature is enabled for Node 10+. Node 8 doesn't implement the `url` property on the `CallFrame` object, and cannot reliably resolve to the correct locations in the stack. (There may be a workaround, but none that I know at this time.)

This PR adheres to the established style of Rollbar.js and uses `var` declarations, rather than `let` and `const`. ES5 style classes are used. I'm not sure whether `session.post()` supports async/await. Regardless, callbacks are used throughout as shared code that cannot be made async prevents the use of async/await in this code path.

## Type of change

## Related issues

Fixes: ch74679

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/src/server/locals.js
+++ b/src/server/locals.js
@@ -43,9 +43,7 @@ Locals.prototype.initSession = function() {
 }
 
 Locals.prototype.currentLocalsMap = function() {
-  if (Locals.currentErrors.size) {
-    return new Map(Locals.currentErrors);
-  }
+  return new Map(Locals.currentErrors);
 }
 
 Locals.prototype.mergeLocals = function(localsMap, stack, key, callback) {
--- a/src/server/locals.js
+++ b/src/server/locals.js
@@ -3,6 +3,10 @@ var inspector = require('inspector');
 var async = require('async');
 
 function Locals(config) {
+  if (!(this instanceof Locals)) {
+    return new Locals(config);
+  }
+
   this.config = config;
 
   this.initSession();
--- a/src/server/locals.js
+++ b/src/server/locals.js
@@ -4,37 +4,43 @@ var async = require('async');
 
 function Locals(config) {
   this.config = config;
-  this.errorCounter = 0;
-  this.currentErrors = new Map();
 
   this.initSession();
 }
 
 Locals.prototype.initSession = function() {
-  this.session = new inspector.Session();
-  this.session.connect();
+  if (Locals.session) { return; }
 
-  this.session.on('Debugger.paused', ({ params }) => {
+  Locals.session = new inspector.Session();
+  Locals.session.connect();
+  Locals.currentErrors = new Map();
+
+  Locals.session.on('Debugger.paused', ({ params }) => {
     if (params.reason == 'promiseRejection' || params.reason == 'exception') {
       var key = params.data.description;
-      this.currentErrors.set(key, params);
+      Locals.currentErrors.set(key, params);
 
-      if (this.currentErrors.size > 4) {
-        var firstKey = this.currentErrors.keys()[0];
-        this.currentErrors.delete(firstKey);
+      // Set the max size of the current errors array.
+      // The value should be large enough to preserve each of the errors in
+      // the current cause chain.
+      var CURRENT_ERRORS_MAX_SIZE = 4;
+
+      if (Locals.currentErrors.size > CURRENT_ERRORS_MAX_SIZE) {
+        var firstKey = Locals.currentErrors.keys()[0];
+        Locals.currentErrors.delete(firstKey);
       }
     }
   });
 
-  this.session.post('Debugger.enable', (_err, _result) => {
-    this.session.post('Debugger.setPauseOnExceptions', { state: 'all'}, (_err, _result) => {
+  Locals.session.post('Debugger.enable', (_err, _result) => {
+    Locals.session.post('Debugger.setPauseOnExceptions', { state: 'all'}, (_err, _result) => {
     });
   });
 }
 
 Locals.prototype.currentLocalsMap = function() {
-  if (this.currentErrors.size) {
-    return new Map(this.currentErrors);
+  if (Locals.currentErrors.size) {
+    return new Map(Locals.currentErrors);
   }
 }
 
@@ -43,18 +49,22 @@ Locals.prototype.mergeLocals = function(localsMap, stack, key, callback) {
 
   try {
     var localParams = localsMap.get(key);
-    matchedFrames = this.matchFrames(localParams, stack.slice().reverse());
+
+    // If a mapping isn't found return success without mapped locals.
+    if (!localParams) {
+      return callback(null);
+    }
+
+    matchedFrames = matchFrames(localParams, stack.slice().reverse());
   } catch (e) {
     return callback(e);
   }
 
-  this.getLocalScopesForFrames(matchedFrames, function(err) {
-    callback(err);
-  });
+  getLocalScopesForFrames(matchedFrames, callback);
 }
 
 // Finds frames in localParams that match file and line locations in stack.
-Locals.prototype.matchFrames = function(localParams, stack) {
+function matchFrames(localParams, stack) {
   var matchedFrames = [];
   var localIndex = 0, stackIndex = 0;
   var stackLength = stack.length;
@@ -63,7 +73,7 @@ Locals.prototype.matchFrames = function(localParams, stack) {
 
   for (; stackIndex < stackLength; stackIndex++) {
     while (localIndex < callFramesLength) {
-      if (this.firstFrame(localIndex, stackIndex) || this.matchedFrame(callFrames[localIndex], stack[stackIndex])) {
+      if (firstFrame(localIndex, stackIndex) || matchedFrame(callFrames[localIndex], stack[stackIndex])) {
         matchedFrames.push({
           stackLocation: stack[stackIndex],
           callFrame: callFrames[localIndex]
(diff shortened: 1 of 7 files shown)


src/server/locals.js
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  return matchedFrames;
+}
+
+Locals.prototype.firstFrame = function(localIndex, stackIndex) {
+  return !localIndex && !stackIndex;
+}
+
+Locals.prototype.matchedFrame = function(callFrame, stackLocation) {
```
eudaimos, line 87: same comment about this function as the above for `firstFrame` - this function isn't making use of `this` - does it need to be exported on the `Locals.prototype`?

src/server/locals.js
```diff
@@ -0,0 +1,161 @@
...
+  var callFrameLine = callFrame.location.lineNumber + 1;
+  var callFrameColumn = callFrame.location.columnNumber;
+
+  return callFrameUrl === position.source &&
+    callFrameLine === position.line &&
+    callFrameColumn === position.column;
+}
+
+Locals.prototype.getLocalScopesForFrames = function(matchedFrames, callback) {
+  async.each(matchedFrames, this.getLocalScopeForFrame.bind(this), function (err) {
```
eudaimos, line 108: seems that you can pass `callback` as the thrid arg to `async.each` rather than having the wrapping function - is there a reason for the extra wrapping function?

seems it will be called no matter the outcome the way it's written

src/server/locals.js
```diff
@@ -0,0 +1,161 @@
...
+  // lineNumber is zero indexed, so offset it.
+  var callFrameLine = callFrame.location.lineNumber + 1;
+  var callFrameColumn = callFrame.location.columnNumber;
+
+  return callFrameUrl === position.source &&
+    callFrameLine === position.line &&
+    callFrameColumn === position.column;
+}
+
+Locals.prototype.getLocalScopesForFrames = function(matchedFrames, callback) {
```
eudaimos, line 107: does this function need to be attached to the `Locals.prototype` and exported? I don't see it used outside of the `mergeLocals` body which does need to be exported (public).

As for the `bind` call to this, that could be done by `mergeLocals` when passing maybe? Just a thought on trying to expose less of these internal methods if they're not needed externally

src/server/locals.js
```diff
@@ -0,0 +1,161 @@
...
+  });
+}
+
+Locals.prototype.getLocalScopeForFrame = function(matchedFrame, callback) {
+  var scopes = matchedFrame.callFrame.scopeChain;
+
+  var _this = this;
+  for (var i = 0; i < scopes.length; i++) {
+    var scope = scopes[i]
+    if (scope.type === 'local') {
```
eudaimos, line 119: why not filter before the loop using a `scopes.filter(scope => scope.type === 'local')`

src/server/locals.js
```diff
@@ -0,0 +1,161 @@
...
+  switch (local.value.type) {
+    case 'undefined': value = 'undefined'; break;
+    case 'object': value = this.getObjectValue(local); break;
+    default: value = local.value.value; break;
+  }
+
+  return value;
+}
+
+Locals.prototype.getObjectValue = function(local) {
```
eudaimos, line 149: does this need to be exported as attached to the `Locals.prototype`?

src/server/locals.js
```diff
@@ -0,0 +1,161 @@
...
+          matchedFrame.stackLocation.locals[local.name] = _this.getLocalValue(local);
+        }
+
+        callback(null);
+      });
+    }
+  }
+}
+
+Locals.prototype.getLocalValue = function(local) {
```
eudaimos, line 137: since the only ref to `this` is for the `this.getObjectValue` call, does this need to be exported on the `Locals.prototype`?

src/server/locals.js
```diff
@@ -0,0 +1,161 @@
+/* globals Map */
+var inspector = require('inspector');
+var async = require('async');
+
+function Locals(config) {
```
eudaimos, line 5: defining classes like this the old school JS way has the problem of possible function call without instantiation, i.e. missing the `new` keyword by just calling `Locals(…)` as a function. To prevent this from doing harm, a pattern emerged for checking how a constructor is called like this:

```javascript
function Locals(config) {
  if (!(this instanceof Locals)) {
    return new Locals(config);
  }
  // rest of constructor logic
}
```

eudaimos (commented):
@waltjones 

Overall I made comments/asked questions about the implementation and public api of `Locals`

Also wondering, should there be unit tests for the `Locals` API?
The tests I see seem to be routed through the rollbar SDK

eudaimos (changes requested):
I never seem to get this right - requested changes are comments in previous "Comment" and submitting this now as a Review for the process to work

src/server/locals.js
waltjones: Updated to use `find` and expect one local scope.

src/server/locals.js
```diff
@@ -0,0 +1,161 @@
...
+Locals.prototype.initSession = function() {
+  this.session = new inspector.Session();
+  this.session.connect();
+
+  this.session.on('Debugger.paused', ({ params }) => {
+    if (params.reason == 'promiseRejection' || params.reason == 'exception') {
+      var key = params.data.description;
+      this.currentErrors.set(key, params);
+
+      if (this.currentErrors.size > 4) {
```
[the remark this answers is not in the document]
waltjones: Constant and explanatory comment added.

src/server/locals.js
```diff
@@ -0,0 +1,161 @@
...
+
+  try {
+    var localParams = localsMap.get(key);
+    matchedFrames = this.matchFrames(localParams, stack.reverse());
+  } catch (e) {
+    console.log(e);
+    return callback(e);
+  }
+
+  this.getLocalScopesForFrames(matchedFrames, function(err) {
```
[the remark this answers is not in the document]
waltjones: Good catch. Each of these cases have been condensed now.

src/server/locals.js
```diff
@@ -0,0 +1,161 @@
...
+      } else {
+        localIndex++;
+      }
+    }
+  }
+
+  return matchedFrames;
+}
+
+Locals.prototype.firstFrame = function(localIndex, stackIndex) {
```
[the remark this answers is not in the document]
waltjones: Non-interface functions are all private now.

src/server/locals.js
waltjones: Added.

waltjones:
Major updates are:
* The inspector session is now a singleton at `Locals.session`. This makes more sense, since multiple Rollbar.js instances should not open multiple sessions. This has the side effect of greatly reducing the need to pass `this` around since the main need for it was when calling `session.post()`.
* Unit tests have been added, which allows exercising some conditions not produced when the actual inspector API is called.

christopherseaman:
Update?

@eudaimos @waltjones

waltjones:
Just waiting on approval or further feedback.

bishopb (approved):

src/server/locals.js
```diff
@@ -0,0 +1,173 @@
...
+
+  Locals.session.on('Debugger.paused', ({ params }) => {
+    if (params.reason == 'promiseRejection' || params.reason == 'exception') {
+      var key = params.data.description;
+      Locals.currentErrors.set(key, params);
+
+      // Set the max size of the current errors array.
+      // The value should be large enough to preserve each of the errors in
+      // the current cause chain.
+      var CURRENT_ERRORS_MAX_SIZE = 4;
```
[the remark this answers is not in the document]
waltjones: For now, the codebase is standardized on `var` rather than `let`/`const`, though I think there are a few `const` instances that have been merged in the past. My preference is to stay consistent until the decision is made to move away from `var`.

The value itself shouldn't need to be customized. That's the intent anyway. If there were a demonstrated need (not just for it to be larger, but for it to be larger only in some cases), it could be parameterized then. It's much easier to add to the interface, than to remove or change something.

src/server/locals.js
```diff
@@ -0,0 +1,173 @@
...
+    Locals.session.post('Debugger.setPauseOnExceptions', { state: 'all'}, (_err, _result) => {
+    });
+  });
+}
+
+Locals.prototype.currentLocalsMap = function() {
+  if (Locals.currentErrors.size) {
+    return new Map(Locals.currentErrors);
+  }
+}
```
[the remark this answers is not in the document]
waltjones: Updated to always return a map. I've kept the check at: https://github.com/rollbar/rollbar.js/blob/aaafe97927a0c0741300725330269f98dd84985c/src/server/parser.js#L313 

because `item.localsMap` will be undefined when locals is disabled.

src/server/locals.js
```diff
@@ -0,0 +1,173 @@
...
+
+Locals.prototype.mergeLocals = function(localsMap, stack, key, callback) {
```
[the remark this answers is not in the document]
waltjones: Addressed above.

PR #914, opened 2021-01-12
Enable depth of traversal for node local vars

waltjones:
## Description of the change

Follow up PR to https://github.com/rollbar/rollbar.js/pull/902 adds traversal of objects and arrays. Capture is performed to the specified depth. Options are also provided to limit the maximum number of object properties and array members captured. 

### Options and defaults

Locals capture may be enabled by setting `options.locals = true` to enable with default options, or with an object containing overrides:
```javascript
options.locals = {
  depth: 2,
  maxProperties: 15,
  maxArray: 10
}
```

`depth` sets the depth to which objects and arrays are expanded. The previous PR supported depth = 0, which captures the value of primitive types, and displays the type of all non-primitives. This is consistent with the example in the Rollbar API doc.

**Example output, depth = 0**
```javascript
locals: {
  foo: "bar",
  res: "<ServerResponse object>", 
  req: "<IncomingMessage object>", 
  next: "<Function object>"
}
``` 

The default is depth = 1, which shows the properties of objects and arrays assigned to local variables.
 
**Example output, depth = 1, (maxProperties = 5 to keep the examples small)**
```javascript
locals: {
  foo: "bar",
  res: {
    _header: null, 
    writable: true,
    _maxListeners: "undefined", 
    _last: false, 
    outputCallbacks: "<Array object>",
  }, 
  req: {
    _readableState: "<ReadableState object>", 
    _eventsCount: 1, 
    baseUrl: "", 
    query: "<Object object>",
    rawHeaders: "<Array object>",
  }, 
  next: "<Function object>" // Function objects are not expanded, see notes below.
}
``` 

Larger values for depth are permitted at the cost of larger payloads.

**Example output, depth = 2 (maxProperties = 5 to keep the examples small)**
```javascript
locals: {
  foo: "bar",
  res: {
    _header: null, 
    writable: true,
    _maxListeners: "undefined", 
    _last: false, 
    outputCallbacks: [
      0, 
      "<Array object>"
    ],
  }, 
  req: {
    _readableState: {
      defaultEncoding: "utf8", 
      length: 0, 
      reading: false, 
      resumeScheduled: false, 
      highWaterMark: 16384, 
      buffer: "<BufferList object>",
    }, 
    _eventsCount: 1, 
    baseUrl: "", 
    query: {}, // empty object, but we didn't detect that with depth = 1 above
    rawHeaders: [
      "Host", 
      "localhost:3000", 
      "Connection", 
      "keep-alive", 
      "Cache-Control"
    ],
  }, 
  next: "<Function object>" // Function objects are not expanded, see notes below.
}
```

`maxProperties` limits the number of properties captured for objects that are traversed and expanded. The default is maxProperties = 30, which can be overriden as needed.

`maxArray` limits the number of members captured for arrays that are traversed and expanded. The default is maxArray = 5, which can be overriden as needed.

### Notes

When objects are expanded, we lose the ability to communicate their type. The information is still available at runtime, but the payload doesn't provide a place to put it. This is effectively by design, and no better or worse than locals support in other SDKs. Annotating the name was considered, e.g. `"foo1 <Foo>": { bar: baz }`. But that key is also used for scrubbing. Altering these keys risks unintended scrub failures.

When functions are assigned as locals, we only show the class name `<Function>` or `<AsyncFunction>`. The function signature and full body are available at runtime, and it would be possible to do something like `<Function  foo(bar)>` in a later PR if there's interest.

Another possible improvement would be to optionally skip locals for stack frames belonging to dependencies. This could be considered in a later PR if there's interest.

## Type of change

## Related issues

Fixes ch75361

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/src/server/locals.js
+++ b/src/server/locals.js
@@ -1,13 +1,28 @@
 /* globals Map */
 var inspector = require('inspector');
 var async = require('async');
+var _ = require('../utility');
 
-function Locals(config) {
+// It's helpful to have default limits, as the data expands quickly in real environments.
+// depth = 1 is  enough to capture the members of top level objects and arrays.
+// maxProperties limits the number of properties captured from non-array objects.
+// When this value is too small, relevant values for debugging are easily omitted.
+// maxArray applies to array objects, which in practice may be arbitrarily large,
+// yet for debugging we usually only care about the pattern of data that is established,
+// so a smaller limit is usually sufficient.
+var DEFAULT_OPTIONS = {
+  depth: 1,
+  maxProperties: 30,
+  maxArray: 5
+}
+
+function Locals(options) {
   if (!(this instanceof Locals)) {
-    return new Locals(config);
+    return new Locals(options);
   }
 
-  this.config = config;
+  options = _.isType(options, 'object') ? options : {};
+  this.options = _.merge(DEFAULT_OPTIONS, options);
 
   this.initSession();
 }
@@ -62,7 +77,7 @@ Locals.prototype.mergeLocals = function(localsMap, stack, key, callback) {
     return callback(e);
   }
 
-  getLocalScopesForFrames(matchedFrames, callback);
+  getLocalScopesForFrames(matchedFrames, this.options, callback);
 }
 
 // Finds frames in localParams that match file and line locations in stack.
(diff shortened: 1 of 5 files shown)


src/server/locals.js
```diff
@@ -135,35 +151,102 @@ function getLocalScopeForFrame(matchedFrame, callback) {
...
+    if (_.typeName(localsObject) === 'array') {
+      localsObject.push(value);
+    } else {
+      localsObject[local.name] = value;
+    }
+    callback(null);
   }
 
-  return value;
+  if(!local.value) {
```
bishopb, line 186: Nit: there's a mix of styles in the PR. Some are `if(` while others are `if (`, the latter seeming to dominate. Ideally, all would have the same style.

src/server/locals.js
```diff
@@ -135,35 +151,102 @@ function getLocalScopeForFrame(matchedFrame, callback) {
...
+
+function getObjectValue(local, options, depth, callback) {
+  if(!local.value.objectId) {
+    if ('value' in local.value) {
+      // Treat as immediate value. (Known example is `null`.)
+      return callback(null, local.value.value);
+    }
+  }
+
+  if (!depth) {
```
bishopb, line 219: Nit: `depth` is named like an integer, but is treated like a boolean. Can we clarify this with `0 === depth`? (Yoda or not, identical or equality; just intent is to make it look like an integer in the conditional.)

bishopb (approved):

waltjones:
@bishopb Good eye, thank you.

Issue #930, opened 2021-02-24
TypeError: t.headers.get is not a function. (In 't.headers.get("Content-Type")', 't.headers.get' is undefined)


dseeker:
We're using an unusual setup with a hybrid iOS application running a compiled js file inside a native iOS app.
Seeing the user's report a lot of them are having an issue with the rollbar lib and I can't quite pinpoint why that's the case. 
There's is probably some errors trying to get notified but it's failing after this line.

Would it be possible to add a check before executing headers.get()? I'm not sure how important setting the json type is, we could have some default value in case getting that header fails

on this line
https://github.com/rollbar/rollbar.js/blob/2cf9c2b85668d1f00314468a6dbbd84dd2583f13/src/browser/telemetry.js#L388

maybe something like this?
```js
metadata.response_content_type = (resp.headers && typeof resp.headers.get === 'function') ? resp.headers.get('Content-Type') : '';
``` 

please let me know if there is a better way to debug this. I can't really reproduce it locally, I'm just seeing this in user's rollbar report

waltjones:
Yes, I think we can do something here.

waltjones:
@dseeker I hoped this could be accommodated while leaving network telemetry enabled. Instead I think it will be necessary to disable in your Rollbar config:

```javascript
config.autoInstrument = { network: false }
```
This should solve the xhr/fetch related errors.

See https://docs.rollbar.com/docs/rollbarjs-telemetry for more details on telemetry support.

PR #932, opened 2021-03-05
Add check for missing headers object

waltjones:
## Description of the change

There are compiled environments that partially support the fetch API. More specifically, the `Response` object either doesn't implement the `headers` property or the `Headers` object doesn't implement the `get()` method.

This PR adds a guard around the access to to `response.headers.get()`. When headers can't be retrieved, it is safe to leave `metadata.response_content_type` undefined, which is what this PR does.

There are other places `response.headers` is used, but only if the telemetry option `networkResponseHeaders` is set. When targeting an environment that doesn't implement this part of the fetch API, the `networkResponseHeaders` option should be left unset.

Testing:
The conditions for the negative case (no headers object) is difficult to replicate, as `Response` and `Headers` are both implemented natively on Chrome and can't be stubbed or modified as far as I know.

The test for the positive case, specifically the test that shows scrubbing was performed on the response body, is currently blocked on a Headless Chrome compatibility issue. This was tested manually and passed. 

## Type of change

## Related issues

Fixes: https://github.com/rollbar/rollbar.js/issues/930
ch80956

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/src/browser/telemetry.js
+++ b/src/browser/telemetry.js
@@ -385,7 +385,9 @@ Instrumenter.prototype.instrumentNetwork = function() {
         return orig.apply(this, args).then(function (resp) {
           metadata.end_time_ms = _.now();
           metadata.status_code = resp.status;
-          metadata.response_content_type = resp.headers.get('Content-Type');
+          if (resp.headers && typeof resp.headers.get === 'function') {
+            metadata.response_content_type = resp.headers.get('Content-Type');
+          }
           var headers = null;
           if (self.autoInstrument.networkResponseHeaders) {
             headers = self.fetchHeaders(resp.headers, self.autoInstrument.networkResponseHeaders);


bxsx:
@waltjones great work! No comment to the changes they look good to me. Just a question regarding this decision:

>There are other places response.headers is used, but only if the telemetry option networkResponseHeaders is set. When targeting an environment that doesn't implement this part of the fetch API, the networkResponseHeaders option should be left unset.

I'm a bit worried about fragility of this. Is there any safeguard implemented to not allow to set the `networkResponseHeaders` for such environments?
If not, I would rather wrap all of the `response.headers` or add a comment.

waltjones:
This telemetry code is specifically in the browser package for Rollbar.js, so the tradeoff for the extra defensive code is a larger browser package size. To my knowledge, no browsers that implement `fetch` fail to fully support it in the way described here. We'd be adding the extra code size for all clients, for a condition we don't expect to see in browsers. Meanwhile, for custom compiled targets like the one in https://github.com/rollbar/rollbar.js/issues/930, the solution is to not enable `networkResponseHeaders` which is off by default.

Are you suggesting to add a comment in each place where `response.headers` is referenced?

bxsx:
> This telemetry code is specifically in the browser package for Rollbar.js, so the tradeoff for the extra defensive code is a larger browser package size. To my knowledge, no browsers that implement `fetch` fail to fully support it in the way described here. We'd be adding the extra code size for all clients, for a condition we don't expect to see in browsers. Meanwhile, for custom compiled targets like the one in #930, the solution is to not enable `networkResponseHeaders` which is off by default.

Fair enough!

> Are you suggesting to add a comment in each place where `response.headers` is referenced?

I was thinking about a comment around `networkResponseHeaders` default option or setter for the particular option (I don't think it exists).
https://github.com/rollbar/rollbar.js/blob/1d59d0fddce53b604aeb97dcedc8a022b0903d35/src/browser/telemetry.js#L6-L21

bxsx:
> Maybe what we'd want would be something more visible to users of the config options, like in the SDK doc.

I agree, this should be described in the SDK docs such as https://docs.rollbar.com/docs/rollbarjs-telemetry
This is definitely a first place to make a comment.

However, if we go along with the PR something like a comment below could be appreciated due to fragility and possibility of the flaw. I'm also OK to accept only the doc update if you don't agree this suggestion.

```javascript
var defaults = { 
   network: true, 
   networkResponseHeaders: false, 
   networkResponseBody: false, 
   networkRequestHeaders: false,  // NOTE: Do NOT enable for custom built browsers
   networkRequestBody: false, 
```
> But either way, this highlights a flaw in the original PR.
> 
> The possibility of some platforms breaking standard or required browser interfaces has always been there. And there are at least dozens of ways a partially implemented platform could omit a requirement that rollbar.js depends on. It appears that by making a single exception, as in this PR, we open the door to responsibility for all the different ways a custom platform might omit or mis-implement required interfaces.

I agree, however I didn't want to make a strong statement about it myself. IMHO this is an issue with JS and web browsers in general.

> 
> Rather than make the one exception in this PR, it would be better to tell people to just turn off network telemetry for this use case.

I'm OK with closing this PR without merge and only adding a suitable information in the Telemetry documentation. If this is what you mean here. One of my principal engineering rule is `Special cases aren't special enough to break the rules. Although practicality beats purity.` :)

PR #934, opened 2021-03-11
Release v2.21.1

waltjones:
## Description of the change

Release v2.21.1, version bump and build artifacts

## Type of change
- [x] Release

## Related issues

ch81339

### Development

- [x] Lint rules pass locally

### Code review 


dist/rollbar.js
```diff
@@ -451,8 +451,8 @@ function makeUnhandledStackInfo(
   };
   location.func = errorParser.guessFunctionName(location.url, location.line);
   location.context = errorParser.gatherContext(location.url, location.line);
-  var href = document && document.location && document.location.href;
-  var useragent = window && window.navigator && window.navigator.userAgent;
+  var href = typeof document !== 'undefined' && document && document.location && document.location.href;
```
pawelsz-rb, line 454: I think document has to be checked first

dist/rollbar.js
```diff
@@ -451,8 +451,8 @@ function makeUnhandledStackInfo(
...
+  var useragent = typeof window !== 'undefined' && window && window.navigator && window.navigator.userAgent;
```
pawelsz-rb, line 455: same as above

dist/rollbar.js
waltjones: This is a build artifact for a previously merged PR, but aside from that, I'm curious. What would the suggested change look like?

pawelsz-rb: var href = document && typeof document !== 'undefined' && document.location && document.location.href;

it makes a difference for python when it's evaluated first, maybe it does not matter for js.

waltjones: It does matter for js. That code will throw a document is not defined exception.

I'm curious again though. In python when document is undefined it will evaluate `true`? If not then`typeof document !== 'undefined'` is superfluous. It will never be reached with undefined.

pawelsz-rb (approved):

dist/rollbar.js
bxsx: @waltjones there is no `undefined` state in Python, so you will get `NameError` exception regarding the undefined object.

So, in Python, you have to be sure an object (id reference to be precious) is defined before you reference to it.

bxsx: I wouldn't say so.
You can't verify if object/id is (un)defined in Python's frontend. There are some workarounds like try/except or by inspecting internal dictionaries etc. However, there is no official/recommended way to do this.

In my opinion, JS is closer to Ruby or Perl here.

waltjones: I meant roughly the same in that you cannot do `if (foo)` in either language when `foo` is undefined.

PR #936, opened 2021-03-19
fix: make rollbar compatible with webpack

andrei-bitca-dc:
## Description of the change

Rollbar is not working when using `@vercel/ncc` to build. It points to the project package.json, instead of rollbar's one, therefore getting:

```
TypeError: Cannot read property 'server' of undefined at Object.3029 (/var/webpack:/node_modules/rollbar/src/server/rollbar.js:659:1)
```

## Type of change

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/src/server/rollbar.js
+++ b/src/server/rollbar.js
@@ -1,7 +1,8 @@
 var util = require('util');
 var os = require('os');
+var path = require('path');
 
-var packageJson = require('../../package.json');
+var packageJson = require(path.join(__dirname, '../../package.json'));
 var Client = require('../rollbar');
 var _ = require('../utility');
 var API = require('../api');


src/server/rollbar.js
```diff
@@ -1,7 +1,8 @@
 var util = require('util');
 var os = require('os');
+var path = require('path');
 
-var packageJson = require('../../package.json');
+var packageJson = require(path.join(__dirname, '../../package.json'));
```
waltjones, line 5: Why does the path only resolve incorrectly here, and not for any of the following relative paths?

andrei-bitca-dc: I think because it's the `package.json` file and it's not included in the dist.

andrei-bitca-dc: However, using `path.join` for dynamic imports won't work with `esbuild`, as it doesn't support dynamic imports yet (https://github.com/evanw/esbuild/issues/700).

PR #943, opened 2021-04-16
Node.js telemetry - logging and http(s) request

waltjones:
## Description of the change

Enables telemetry in Node.js for logging and http(s) request.

**Logging:** 
* Hooks `process.stdout` and `process.stderr`.
* Records the log string and log level.

**Network:**
* Hooks `http.request` and `https.request`.
* Network capture records method/verb, url, status code, start/end time in ms.
* Non-HTTP network errors (connection errors, DNS errors, etc.) are recorded.
* When header capture is enabled, headers are scrubbed based on `scrubHeaders` settings. 

**Config options are consistent with browser js telemetry options:**
```
{
  autoInstrument: true // enable logging and network capture, without network headers
}
```
```
{
  autoInstrument: {
    network: true, // enable/disable network and log capture independently
    log: true 
  }
```
```
{
  autoInstrument: {
    network: true, 
    networkRequestHeaders: true, // Capture request and/or response headers
    networkResponseHeaders: true 
  }
```

**Future work (items not covered in this PR):**
* http/2 streams
* Capture request/response bodies
* Non-HTTP network: sockets, udp, etc

## Type of change

## Related issues

ch82559

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/src/server/rollbar.js
+++ b/src/server/rollbar.js
@@ -12,6 +12,7 @@ var urllib = require('url');
 var jsonBackup = require('json-stringify-safe');
 
 var Telemeter = require('../telemetry');
+var Instrumenter = require('./telemetry');
 var transforms = require('./transforms');
 var sharedTransforms = require('../transforms');
 var sharedPredicates = require('../predicates');
@@ -40,6 +41,8 @@ function Rollbar(options, client) {
   var api = new API(this.options, transport, urllib, truncation, jsonBackup);
   var telemeter = new Telemeter(this.options)
   this.client = client || new Client(this.options, api, logger, telemeter, 'server');
+  this.instrumenter = new Instrumenter(this.options, this.client.telemeter, this);
+  this.instrumenter.instrument();
   if (this.options.locals) {
     this.locals = initLocals(this.options.locals, logger);
   }
@@ -670,7 +673,8 @@ Rollbar.defaultOptions = {
   captureIp: true,
   captureLambdaTimeouts: true,
   ignoreDuplicateErrors: true,
-  scrubRequestBody: true
+  scrubRequestBody: true,
+  autoInstrument: false
 };
 
 module.exports = Rollbar;
--- a/src/server/telemetry.js
+++ b/src/server/telemetry.js
@@ -149,6 +149,12 @@ Instrumenter.prototype.instrumentConsole = function() {
   }, this.replacements, 'log');
 };
 
+// TODO: These helpers are duplicated in src/browser/telemetry.js,
+// and may be candidates for extraction into a shared module.
+// It is recommended that before doing so, the author should allow
+// for more telemetry types to be implemented for the Node target
+// to ensure that the implementations of these helpers don't diverge.
+// If they do diverge, there's little point in the shared module.
 function replace(obj, name, replacement, replacements, type) {
   var orig = obj[name];
   obj[name] = replacement(orig);
--- a/src/server/telemetry.js
+++ b/src/server/telemetry.js
@@ -1,0 +1,168 @@
+var http = require('http');
+var https = require('https');
+var _ = require('../utility');
+var urlHelpers = require('./telemetry/urlHelpers');
+
+var defaults = {
+  network: true,
+  networkResponseHeaders: false,
+  networkRequestHeaders: false,
+  log: true
+};
+
+function Instrumenter(options, telemeter, rollbar) {
+  this.options = options;
+  var autoInstrument = options.autoInstrument;
+  if (options.enabled === false || autoInstrument === false) {
+    this.autoInstrument = {};
+  } else {
+    if (!_.isType(autoInstrument, 'object')) {
+      autoInstrument = defaults;
+    }
+    this.autoInstrument = _.merge(defaults, autoInstrument);
+  }
+  this.telemeter = telemeter;
+  this.rollbar = rollbar;
+  this.diagnostic = rollbar.client.notifier.diagnostic;
+  this.replacements = {
+    network: [],
+    log: []
+  };
+}
+
+Instrumenter.prototype.configure = function(options) {
+  this.options = _.merge(this.options, options);
+  var autoInstrument = options.autoInstrument;
+  var oldSettings = _.merge(this.autoInstrument);
+  if (options.enabled === false || autoInstrument === false) {
+    this.autoInstrument = {};
+  } else {
+    if (!_.isType(autoInstrument, 'object')) {
+      autoInstrument = defaults;
+    }
+    this.autoInstrument = _.merge(defaults, autoInstrument);
+  }
+  this.instrument(oldSettings);
+};
+
+Instrumenter.prototype.instrument = function(oldSettings) {
+  if (this.autoInstrument.network && !(oldSettings && oldSettings.network)) {
+    this.instrumentNetwork();
+  } else if (!this.autoInstrument.network && oldSettings && oldSettings.network) {
+    this.deinstrumentNetwork();
+  }
+
+  if (this.autoInstrument.log && !(oldSettings && oldSettings.log)) {
+    this.instrumentConsole();
+  } else if (!this.autoInstrument.log && oldSettings && oldSettings.log) {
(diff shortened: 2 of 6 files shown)


src/server/rollbar.js
```diff
@@ -12,6 +12,7 @@ var urllib = require('url');
 var jsonBackup = require('json-stringify-safe');
 
 var Telemeter = require('../telemetry');
+var Instrumenter = require('./telemetry');
```
bishopb, line 15: Nit: this _looks_ like a typo, since the line immediately before references the parent directory while this references the current directory. I'd suggest - in a separate PR - renaming the files to better match their intent; perhaps:

* `../telemetry` => `../telemeter`
* `./telemetry` => `./instrumenter`

src/server/telemetry.js
```diff
@@ -0,0 +1,168 @@
...
+  network: true,
+  networkResponseHeaders: false,
+  networkRequestHeaders: false,
+  log: true
+};
+
+function Instrumenter(options, telemeter, rollbar) {
+  this.options = options;
+  var autoInstrument = options.autoInstrument;
+  if (options.enabled === false || autoInstrument === false) {
```
bishopb, line 16: Nit: do we want to support truthy values, rather than strict boolean? I can see someone potentially doing something like this: `options.enabled = someThing.length`.

src/server/telemetry.js
```diff
@@ -0,0 +1,168 @@
...
+    this.autoInstrument = {};
+  } else {
+    if (!_.isType(autoInstrument, 'object')) {
+      autoInstrument = defaults;
+    }
+    this.autoInstrument = _.merge(defaults, autoInstrument);
```
bishopb, line 22: Nit: I think this would read more clearly as:

```
} else if (_.isType(autoInstrument, 'object')) {
  this.autoInstrument = _.merge(defaults, autoInstrument);
} else {
  this.autoInstrument = defaults;
}
```

src/server/telemetry.js
```diff
@@ -0,0 +1,168 @@
...
+  var autoInstrument = options.autoInstrument;
+  var oldSettings = _.merge(this.autoInstrument);
+  if (options.enabled === false || autoInstrument === false) {
+    this.autoInstrument = {};
+  } else {
+    if (!_.isType(autoInstrument, 'object')) {
+      autoInstrument = defaults;
+    }
+    this.autoInstrument = _.merge(defaults, autoInstrument);
+  }
```
bishopb, line 44: Ditto with my comment above about clarifying the deep nesting here.

src/server/telemetry/urlHelpers.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+  url += options.path || '/';
+
+  return url;
+}
+
+module.exports = {
+  mergeOptions,
+  constructUrl
+};
```
bishopb, line 80: I worry these will desynch. If a vulnerability's discovered upstream, is there a way we can reliably receive and act on CVE?

src/server/telemetry.js
```diff
@@ -0,0 +1,168 @@
...
+}
+
+function restore(replacements, type) {
+  var b;
+  while (replacements[type].length) {
+    b = replacements[type].shift();
+    b[0][b[1]] = b[2];
+  }
+}
+
```
bishopb, line 167: Nit: Even though these are covered by integration tests later, I think unit tests for these would go a long way toward self-documenting how they work together and the intent of the implementation.

bishopb (changes requested):

src/server/rollbar.js
waltjones: I tend to agree. For this PR, I wanted to stick with the current naming convention. Changing it, while keeping the code base as a whole consistent, would mean touching a number of files outside the scope of this PR.

src/server/telemetry.js
waltjones: This does support truthy, and significantly, it also treats `null` or `undefined` as truthy. In other words, omitting the option lets it be enabled by default.

Both the interface spec and the implementation are based on the existing browser version of this feature. Given there's more than one way to do this, I preferred consistency over other possible ideas.

src/server/telemetry.js
waltjones: Almost. 

`_.merge()` returns a clone, so you want to take care to do:
`this.autoInstrument = _.merge(defaults); // acts as a deep clone`

The original version of this code does this, though the revised version makes it more apparent.

src/server/telemetry.js
waltjones: :+1:

src/server/telemetry.js
waltjones: I agree, but they can't be accessed as is without exporting them, and doing so could be considered an interface violation (depending on where one stands on that debate.)

These happen to be duplicated code from browser/telemetry.js and exctracting them into a shared module would eliminate the code duplication and create reasonable access for unit tests. I didn't deem that to be in scope for this PR though.

src/server/telemetry/urlHelpers.js
waltjones: I think they are guaranteed to diverge. The goal though is to produce an options object that matches as well as possible the options that will be generated by Node. What is a better way to accomplish that? (Speaking specifically about `mergeOptions` and `urlToHttpOptions`. `constructUrl` is original code, and therefore not relevant to the topic.)

src/server/telemetry.js
pawelsz-rb: I usually try to avoid if/else if possible, so to me, the original way is cleaner if it's doing the same thing. (it's just my preference - it seems to look cleaner, but I fine with both ways)

src/server/telemetry.js
bishopb: Ok, then I think commenting this is duplicated code and an opportunity to refactor later with unit tests suffices to help future maintainers.

src/server/telemetry/urlHelpers.js
bishopb: Ideally, I suppose, would be rewriting our logic to make the need for these disappear. I don't think that's possible in this case, though. A comment pointing to the original source whence it came would at least allow future maintainers to manually cross-check if changes had been made.

src/server/telemetry.js
bishopb: Ah, right. So perhaps most self-evident:

```
} else if (_.isType(autoInstrument, 'object')) {
  this.autoInstrument = _.merge(defaults, autoInstrument);
} else {
  this.autoInstrument = _.cloneDeep(defaults);
}
```

Or, with @pawelsz-rb comment, perhaps this is the most self-documenting:

```
} else {
  this.autoInstrument = _.merge(defaults, _.isType(autoInstrument, 'object') ? autoInstrument : {});
}
```

In either case, I feel readability is enhanced and at the same time the extra not and the potential no-op of merging `defaults` with itself are removed.

src/server/telemetry/urlHelpers.js
waltjones: The comments do have that. Reviewers of previous Rollbar PRs have asked that comments not contain github URLs or github specific paths, so the references here just show the relative path within the stated project.

src/server/telemetry.js
waltjones: I think we might be bikeshedding a little bit, but first a couple things.

I enjoy using the ternary operator, but IMO that example isn't that readable. Maybe the nested function calls and multiple appearances of `autoInstrument` contribute to this.

I'm not opposed to `deepClone()` that calls `merge()` with one argument, but I'm not sure this PR is where to do it. `merge()` with a single argument is already used liberally throughout the code base. I'd rather address it at that level than a one-off here.

src/server/telemetry.js
waltjones: Comment added. https://github.com/rollbar/rollbar.js/pull/943/commits/14a69df80c7c805b3cd1824be2a00b4081a17db4

bishopb (approved):
💪🏼

PR #945, opened 2021-05-03
Apply ignoredMessages filter to trace_chain occurrences

waltjones:
## Description of the change

Previously when matching the `ignoredMessages` array to the item, `trace_chain` items were skipped. This PR adds matching for `trace_chain` messages. 

## Type of change

## Related issues

ch78345

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/src/predicates.js
+++ b/src/predicates.js
@@ -142,8 +142,7 @@ function messageIsIgnored(logger) {
       for (i = 0; i < len; i++) {
         rIgnoredMessage = new RegExp(ignoredMessages[i], 'gi');
 
-        var messagesLength = messages.length;
-        for (j = 0; j < messagesLength; j++) {
+        for (j = 0; j < messages.length; j++) {
           messageIsIgnored = rIgnoredMessage.test(messages[j]);
 
           if (messageIsIgnored) {
--- a/src/predicates.js
+++ b/src/predicates.js
@@ -122,9 +122,7 @@ function urlIsOnAList(item, settings, safeOrBlock, logger) {
 
 function messageIsIgnored(logger) {
   return function(item, settings) {
-    var exceptionMessage, i, ignoredMessages,
-        len, messageIsIgnored, rIgnoredMessage,
-        body, traceMessage, bodyMessage;
+    var i, j, ignoredMessages, len, messageIsIgnored, rIgnoredMessage, messages;
 
     try {
       messageIsIgnored = false;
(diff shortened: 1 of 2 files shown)


src/predicates.js
```diff
@@ -134,23 +132,23 @@ function messageIsIgnored(logger) {
...
 
       len = ignoredMessages.length;
       for (i = 0; i < len; i++) {
         rIgnoredMessage = new RegExp(ignoredMessages[i], 'gi');
-        messageIsIgnored = rIgnoredMessage.test(exceptionMessage);
 
-        if (messageIsIgnored) {
-          break;
+        var messagesLength = messages.length;
+        for (j = 0; j < messagesLength; j++) {
```
bxsx, line 146: All local variables are declared in the function header. I would shift the `messageLength` declaration up.

However, the `messageLength` adds no value, and for that I would simplify it to:

```suggestion
        for (j = 0; j < messages.length; j++) {
```

bxsx (approved):
Good job!

I only have a design suggestion here.

The functionality of the `messageIsIgnored` function is much more complex than the name suggests.

Based on the name of the function, I would expect to receive a boolean that simply says if message is ignored or not.

However, it seems to return `true` in additional cases:
```js
if (!ignoredMessages || ignoredMessages.length === 0) {
        return true;      
}
```
I would rather expect an exception to be thrown instead of returning `true` or at least refactor the part by renaming the function name and extracting the functionality to more single-responsible functions.

src/predicates.js
waltjones: Whether to invoke `length` in a for loop is a longstanding debate. (ex:  https://stackoverflow.com/questions/17989270/for-loop-performance-storing-array-length-in-a-variable) I think it would be fine in this case, but opted for the existing code style for this project.

waltjones:
Re. the contract for `messageIsIgnored`, I have kept the behavior identical to before, as it is a user facing behavior. Similarly with the naming of it, it has been named this way for a long time and is one of the standard predicates. With the goal here to support trace chain, further changes seem out of scope.

src/predicates.js
bxsx: Interesting point, @waltjones. However, I wouldn't trust a random StackOverflow thread. Keep in mind that all modern web browsers have implemented JIT and one of the JS JIT principles is loop expansion, so IMHO there is nothing to worry about. Of course, in some random tests, the results may vary. There are also some safe guards around the recompiled code, so in some scenarios it's possible to prove that _"jittable"_ version is _"slower"_. At least that was truth about ~ 10 yrs ago, when we were implementing PyPy JIT based on JavaScript JIT principles.

waltjones: Bart, that's an antagonistic thing to say. I'm not trusting a random thread. I stated clearly that I followed the style of the current project, not a random internet thread.

For example,
https://github.com/rollbar/rollbar.js/blob/v2.21.1/src/predicates.js#L62-L63
and
https://github.com/rollbar/rollbar.js/blob/v2.21.1/src/predicates.js#L102-L103

in the same file as this change.

bxsx: Oh, sorry Walt. I totally agree that it's important to keep consistent style across the project. So, in fact, I would rather stick with your previous version.

Since the PR has already been approved and merged, I just wanted to comment on the general question that was raised in the StackOverflow thread (because I found this debate interesting). I didn't mean to say you're relying on random SO threads. I have your point that this has been debated for a long time and I just wanted to put my two cents into the debate itself. I referred to general quality of approved answers on StackOverflow that sometimes are not the best answers though.

I'm sorry for the misunderstanding!

waltjones: :+1: Thank you, and no worries.

PR #947, opened 2021-05-12
Read timeout from options and pass it to the XMLHttpRequest

albertdaurell:
## Description of the change

From https://github.com/rollbar/rollbar.js/issues/948 we decided to add `maxRetries` and `timeout` parameters to the rollbar config:

```javascript
var rollbarConfig = {
      retryInterval: 10000,
      maxRetries: 3, // Only used if retryInterval is set
      timeout: 5000, // HTTP Requests timeout
      ...
    };
```
## Type of change

## Related issues

https://github.com/rollbar/rollbar.js/issues/948

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/src/browser/transport.js
+++ b/src/browser/transport.js
@@ -111,6 +111,10 @@ function _proxyRequest(json, callback) {
   );
 }
 
+function _isNumber(n) {
+  return !isNaN(parseFloat(n)) && isFinite(n);
+}
+
 function _makeRequest(accessToken, url, method, data, callback, requestFactory, transportOptions) {
   if (typeof RollbarProxy !== 'undefined') {
     return _proxyRequest(data, callback);
@@ -173,7 +177,7 @@ function _makeRequest(accessToken, url, method, data, callback, requestFactory, 
         request.setRequestHeader('X-Rollbar-Access-Token', accessToken);
       }
 
-      if(transportOptions.timeout) {
+      if(_isNumber(transportOptions.timeout)) {
         request.timeout = transportOptions.timeout;
       }
 
--- a/src/browser/transport.js
+++ b/src/browser/transport.js
@@ -32,7 +32,8 @@ Transport.prototype.get = function(accessToken, options, params, callback, reque
 
   var method = 'GET';
   var url = _.formatUrl(options);
-  _makeZoneRequest(accessToken, url, method, null, callback, requestFactory);
+  var transportOptions = _getMakeZoneRequestOptions(options);
+  _makeZoneRequest(accessToken, url, method, null, callback, requestFactory, transportOptions);
 }
 
 Transport.prototype.post = function(accessToken, options, payload, callback, requestFactory) {
@@ -57,7 +58,8 @@ Transport.prototype.post = function(accessToken, options, payload, callback, req
   var writeData = stringifyResult.value;
   var method = 'POST';
   var url = _.formatUrl(options);
-  _makeZoneRequest(accessToken, url, method, writeData, callback, requestFactory);
+  var transportOptions = _getMakeZoneRequestOptions(options);
+  _makeZoneRequest(accessToken, url, method, writeData, callback, requestFactory, transportOptions);
 }
 
 Transport.prototype.postJsonPayload = function (accessToken, options, jsonPayload, callback, requestFactory) {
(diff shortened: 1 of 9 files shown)


src/browser/transport.js
```diff
@@ -100,7 +111,11 @@ function _proxyRequest(json, callback) {
   );
 }
 
-function _makeRequest(accessToken, url, method, data, callback, requestFactory) {
+function _isNumber(n) {
+  return !isNaN(parseFloat(n)) && isFinite(n);
+}
```
waltjones, line 116: This would be better placed in this section of src/utility.js: https://github.com/rollbar/rollbar.js/blob/v2.21.1/src/utility.js#L71-L144

Is there a reason we should allow coercion of strings, rather than requiring the the value to be a number?

Since `isFinite()` also checks for `NaN`, can we write this without `isNaN`?

Maybe:
```
function isFiniteNumber(n) {
  return Number.isFinite(n);
}
```
does everything we want here.

src/browser/transport.js
```diff
@@ -57,7 +58,8 @@ Transport.prototype.post = function(accessToken, options, payload, callback, req
   var writeData = stringifyResult.value;
   var method = 'POST';
   var url = _.formatUrl(options);
-  _makeZoneRequest(accessToken, url, method, writeData, callback, requestFactory);
+  var transportOptions = _getMakeZoneRequestOptions(options);
```
waltjones, line 61: There are a couple other ways to handle this that might be simpler.

It would be OK to just pass the `options` as is, without filtering to just include the timeout. Or it would also be OK to pass the timeout as an argument. There's only one option in this case, so it doesn't necessarily need to be wrapped in an options argument.

I would favor either of those approaches over the extra code to support `_getMakeZoneRequestOptions`. We're always looking for opportunities to keep the browser package smaller, and rebuilding the options object here isn't strictly necessary.

src/browser/transport.js
```diff
@@ -67,24 +69,33 @@ Transport.prototype.postJsonPayload = function (accessToken, options, jsonPayloa
...
-function _makeZoneRequest(accessToken, url, method, data, callback, requestFactory) {
+function _makeZoneRequest(accessToken, url, method, data, callback, requestFactory, transportOptions) {
   var gWindow = ((typeof window != 'undefined') && window) || ((typeof self != 'undefined') && self);
   var currentZone = gWindow && gWindow.Zone && gWindow.Zone.current;
 
   if (currentZone && currentZone._name === 'angular') {
     var rootZone = currentZone._parent;
     rootZone.run(function () {
-      _makeRequest(accessToken, url, method, data, callback, requestFactory);
+      _makeRequest(accessToken, url, method, data, callback, requestFactory, transportOptions);
```
waltjones, line 95: The original code here and on line 98 below would have been better as `_makeRequest(arguments);` as it is just passing through the arguments as is. Since we're touching this code, now would be a good time to make this improvement.

waltjones:
@albertdaurell This is looking good overall. Thanks for your work on this!

src/browser/transport.js
albertdaurell: Agree, moved source code

src/browser/transport.js
albertdaurell: Ok, removed _getMakeZoneRequestOptions and passed only timeout

src/browser/transport.js
albertdaurell: @waltjones Maybe you prefer doing this:

image

Is it ok for you adding "this" for the thisArg? I've checked your source code and it seems "this" is valid although I don't know if you prefer "undefined" ?

waltjones: Yes that's what I intended. Thank you for correcting that.

`this` won't reference `Transport` here, so I think using undefined makes the intent more clear.

albertdaurell: Perfect, applied changes

test/browser.transport.test.js
```diff
@@ -31,54 +32,59 @@ describe('post', function() {
...
-    var requestFactory = requestGenerator(response, 500, true);
+    var requestFactory = new requestGenerator(response, 500, true);
     var callback = function(err, resp) {
       expect(resp).to.not.be.ok();
       expect(err.message).to.match(/Cannot find a method to transport/);
+      expect(requestFactory.request.timeout).to.equal(options.timeout);
       done(resp);
     };
-    t.post(accessToken, options, payload, callback, requestFactory);
+    t.post(accessToken, options, payload, callback, requestFactory.get.bind(requestFactory));
```
waltjones, line 87: In each of these cases, isn't `get` already bound to `requestFactory` since it is defined on the prototype?

waltjones:
@albertdaurell Do you mind also adding the new options to the Typescript definitions? https://github.com/rollbar/rollbar.js/blob/master/index.d.ts#L43-L96

test/browser.transport.test.js
albertdaurell: No, this is not bound, sorry, I'll write a closure maybe it's "clear" than binding...

albertdaurell:
> @albertdaurell Do you mind also adding the new options to the Typescript definitions? https://github.com/rollbar/rollbar.js/blob/master/index.d.ts#L43-L96

@waltjones Of course not! Commit done, please review :)

waltjones (approved):
@albertdaurell Thank you for the PR. :+1:

Issue #955, opened 2021-06-11
rollbar.log calls do not gracefully handle cyclic references in custom object argument


AndrewJDR:
Repro:
1. Paste the code at the end of this issue into a js file
2. Ensure rollbar is installed (`npm install rollbar`)
3. run `ROLLBAR_ACCESS_TOKEN=<token> node <file>.js`

Result: The error is not reported to rollbar. Instead, a new exception is thrown within rollbar's merge function, which is currently not capable of dealing with cyclic references in the custom object passed to it. This error message is shown on the console:
```
Uncaught Exception thrown RangeError: Maximum call stack size exceeded
    at merge (/home/dev/Documents/src/rollbar-cyclic-object-crash/node_modules/rollbar/src/merge.js:1:1)
    at merge (/home/dev/Documents/src/rollbar-cyclic-object-crash/node_modules/rollbar/src/merge.js:44:26)
    at merge (/home/dev/Documents/src/rollbar-cyclic-object-crash/node_modules/rollbar/src/merge.js:44:26)
    at merge (/home/dev/Documents/src/rollbar-cyclic-object-crash/node_modules/rollbar/src/merge.js:44:26)
    at merge (/home/dev/Documents/src/rollbar-cyclic-object-crash/node_modules/rollbar/src/merge.js:44:26)
    at merge (/home/dev/Documents/src/rollbar-cyclic-object-crash/node_modules/rollbar/src/merge.js:44:26)
    at merge (/home/dev/Documents/src/rollbar-cyclic-object-crash/node_modules/rollbar/src/merge.js:44:26)
    at merge (/home/dev/Documents/src/rollbar-cyclic-object-crash/node_modules/rollbar/src/merge.js:44:26)
    at merge (/home/dev/Documents/src/rollbar-cyclic-object-crash/node_modules/rollbar/src/merge.js:44:26)
    at merge (/home/dev/Documents/src/rollbar-cyclic-object-crash/node_modules/rollbar/src/merge.js:44:26)
```

Expected result:
Rollbar's merge() function used on the custom object paramter should probably deal with cyclic references. I believe Rollbar's scrubbing is already set up to handle cyclic references, so similar logic can probably be used here on the custom object?

code:
```
const Rollbar = require("rollbar");

const rollbar = new Rollbar({
  accessToken: process.env.ROLLBAR_ACCESS_TOKEN,
  enabled: true,
  environment: "production",
});

process.on('uncaughtException', err => {
[...]

PR #990, opened 2022-02-24
Handle circular references in custom data

waltjones:
## Description of the change

This PR guards against circular references in user data (e.g. custom data object) by cloning the data and removing the circular references.

**Why is this necessary?**
The linked issue report (https://github.com/rollbar/rollbar.js/issues/955) notes that the merge function in Rollbar.js errors on the circular references. But JSON serialization will also fail, even if it didn't fail during merge. The circular reference causes an infinite traversal and stack overflow.

**Why can't this be solved within the merge function?**
That's the first solution I considered. It would have had the advantage of adding circular reference protection everywhere merge is used. However, merge doesn't traverse arrays (except the root, if it is an array), and adding array traversal would change its existing behavior. If there are circular references behind an array, merge will succeed, but JSON serialization will fail.

**Why use deepStrictEqual in the node.js tests?**
`assert.equal`, as was previously used for those tests, asserts an identical object reference, not exactly equal objects. Now that the code path clones the user data, the previous assert fails. As a side note, most cases of `assert.equal` in the node.js tests should really be `assert.deepStrictEqual`.

## Type of change

## Related issues

Fixes https://github.com/rollbar/rollbar.js/issues/955

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/utility.js
+++ b/src/utility.js
@@ -398,6 +398,37 @@ function wrapCallback(logger, f) {
   };
 }
 
+function nonCircularClone(obj) {
+  var seen = [obj];
+
+  function clone(obj, seen) {
+    var value, name, newSeen, result = {};
+
+    try {
+      for (name in obj) {
+        value = obj[name];
+
+        if (value && (isType(value, 'object') || isType(value, 'array'))) {
+          if (seen.includes(value)) {
+            result[name] = 'Removed circular reference: ' + typeName(value);
+          } else {
+            newSeen = seen.slice();
+            newSeen.push(value);
+            result[name] = clone(value, newSeen);
+          }
+          continue;
+        }
+
+        result[name] = value;
+      }
+    } catch (e) {
+      result = 'Failed cloning custom data: ' + e.message;
+    }
+    return result;
+  }
+  return clone(obj, seen);
+}
+
 function createItem(args, logger, notifier, requestKeys, lambdaContext) {
   var message, err, custom, callback, request;
   var arg;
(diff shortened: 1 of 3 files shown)


cyrusradfar (approved):
thank you for the context on the bug.

AndrewJDR:
Hi @waltjones and @cyrusradfar,

Thanks for this, I'm the OP on the #955. I recently updated to this version in a staging environment of a node web service (running nsolid), and I've noticed some lockups at the same time that a rollbar error call is made. nsolid reports long blocking times in rollbar's `clone` function at this same time:
```
ID c89dc3a Process blocked for 292 ms at clone (/web/node_modules/rollbar/src/utility.js:404:17) at clone (/web/node_modules/rollbar/src/utility.js:417:28) at clone (/web/node_modules/rollbar/src/utility.js:417:28) at clone (/web/node_modules/rollbar/src/utility.js:417:28) at clone (/web/node_modules/rollbar/src/utility.js:417:28) at clone (/web/node_modules/rollbar/src/utility.js:417:28) at clone (/web/node_modules/rollbar/src/utility.js:417:28) at clone (/web/node_modules/rollbar/src/utility.js:417:28) at clone (/web/node_modules/rollbar/src/utility.js:417:28) at clone (/web/node_modules/rollbar/src/utility.js:417:28)
```

I don't have perfectly definitive repro yet, but I suspect some type of infinite loop condition. I'm going to try to narrow things down in the coming days, but I wanted to let you know sooner than later. It may not be the best idea to publish this into an npm version yet...

PR #997, opened 2022-03-14
Typescript definition improvement: Added types for payload property

paustint:
## Description of the change

Improve typescript type definition.

Added types for Payload instead of using `object`
Added comments with links to documentation
Added comments to aid in understanding code_version
Added optional Promise return type for transform

**Motivation**:
I spent hours trying to troubleshoot and figure out why items were not being matched up with source maps and the typescript definitions left me confused and unsure about how to configure the payload property because the documentation, at times, is not explicit about where a property lives.

For example, this page https://docs.rollbar.com/docs/source-control#serverroot makes it seem like there is a root level configuration option called `server` (which as far as I can tell that is not the case)
And this example seems to show the `server` property in the root (I guess?) https://docs.rollbar.com/docs/source-maps#can-i-still-set-up-source-linking-when-using-source-maps

Either way, the payload property is well defined(ish) in the docs, so there is no reason it should be `object` in the type definition.

I did add `[property: string]: any;` to ever level in the `payload` property and all nested children so that typescript users can supply anything they want, but will get type completion and jsdoc comments for known properties. I wasn't sure how else the payload is used outside what the docs mention.

## Type of change

- [x] Maintenance
- [ ] New release

## Related issues

> Shortcut stories and GitHub issues (delete irrelevant)

- Fix [SC-]
- Fix #1

### Development

(These are all N/A since as far as I can tell, there are no tests for the Typescript definition file 😳 )

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -40,6 +40,9 @@ declare namespace Rollbar {
     ) => void | Promise<TResult>;
     export type MaybeError = Error | undefined | null;
     export type Level = "debug" | "info" | "warning" | "error" | "critical";
+    /**
+     * {@link https://docs.rollbar.com/docs/rollbarjs-configuration-reference#reference}
+     */
     export interface Configuration {
         accessToken?: string;
         addErrorContext?: boolean;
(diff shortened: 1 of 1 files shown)


waltjones (changes requested):
@paustint These are good improvements. Thanks for the PR!

My only concern is about back compatibility with a few of the types. The ones that concern me most are `payload.person.id` and `payload.client.javascript.code_version`. In both cases, the Rollbar API, back end, and dashboard filtering all happen to work fine when using a number type and I'd like to avoid that kind of breaking change for people who are currently setting a number value. That said, the correct type is string, so I'd like to allow number but mark it deprecated.

Something like:
```
    /**
      * @deprecated number is deprecated for this field
      */
    export type DeprecatedNumber = number;
```
And then
```
id: string | DeprecatedNumber;
```
```
code_version: string | DeprecatedNumber;
```
Let me know if you see a better way to do this.

I noticed the eslint deprecation plugin won't warn on this as I'd have expected. This appears to be a known issue described here. https://github.com/gund/eslint-plugin-deprecation/issues/13

Separately, to answer the question about usage of the payload key in general. The contents of this key are merged to the root of the payload before sending. So, for example, the `payload.server` key does get mapped to a root level `server` key. `payload: { foo: 'bar'}` will put a `foo` key in the root of the payload. 

Adding `[property: string]: any;` to each object is a good call, as people may use the payload key in arbitrary ways.

Issue #963, opened 2021-08-10
captureIp has no effect for Electron apps


shadow-light:
The `captureIp` config param appears to be ignored when there is no `window`, i.e. Node. Which makes sense when you are capturing errors on a node server. But it also stops capturing IP for Electron apps, where it can be useful to be able to know how many users are affected by an issue.

A workaround is to manually set `payload.request.user_ip` to `$remote_ip` or `$remote_ip_anonymize`.

waltjones:
Released in v2.25.2, https://github.com/rollbar/rollbar.js/releases/tag/v2.25.2

PR #1008, opened 2022-03-21
Allow captureIp without a window object

waltjones:
## Description of the change

The user IP address is optionally captured at the Rollbar API. The SDK sends a token in the payload signaling whether the captured IP should be added to the payload. This token is added to the payload `request` object, which is only added if a browser window object is detected. (The other keys in the request object depend on `window.location`.)

Setting this token doesn't rely on the `window` object and should be allowed. This is relevant for Electron apps, and for any other client side environment that doesn't support `window.location`.

This PR allows the request object to be sent with the user IP token, regardless of the presence of `window` or `window.location`. Tests are added for this condition, as well as for generally correct behavior when `config.captureIp` is set. 

## Type of change

## Related issues

Fixes https://github.com/rollbar/rollbar.js/issues/963

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/browser/transforms.js
+++ b/src/browser/transforms.js
@@ -79,20 +79,25 @@ function addBaseInfo(item, options, callback) {
 
 function addRequestInfo(window) {
   return function(item, options, callback) {
-    if (!window || !window.location) {
-      return callback(null, item);
+    var requestInfo = {};
+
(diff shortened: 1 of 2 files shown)


ijsnow (approved):
lgtm

Issue #1010, opened 2022-03-24
Add support for Error cause field


twschiller:
- Node and modern browsers [support an optional "cause" field](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error#rethrowing_an_error_with_a_cause)
- For debugging re-thrown errors, it would be great if rollbar.js would support the nested errors and properly include the causes in the payload/stack trace

cyrusradfar:
@twschiller Can't make promises, atm, butI  have been following the development of the Cause standard that's floating around. 

Is this the proposal that you want us to support: https://github.com/tc39/proposal-error-cause

twschiller:
Thanks for the quick response @cyrusradfar. Yes, that link is for the stage 1 proposal

- IIRC: made it into the language: https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-installerrorcause
- Archived Stage 4 proposal: https://github.com/tc39/proposal-error-cause
- Finished proposal tracking: https://github.com/tc39/proposals/blob/main/finished-proposals.md

fregante:
@cyrusradfar I'd expect the `.cause` chain to be displayed similarly to this:

<img width="809" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/160231759-b7eec939-d3b9-4329-98d7-b07be425518d.png">

waltjones:
Released in 2.25.0. https://github.com/rollbar/rollbar.js/releases/tag/v2.25.0

PR #1012, opened 2022-04-01
Support error cause

waltjones:
## Description of the change

This PR adds support for the Error cause option enabled in Node 16.9, Chrome 93, and other browsers. (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error#browser_compatibility)

This PR leaves the existing `Error.nested` functionality for back compatibility with existing Rollbar.js usage and with earlier Node and browser versions. When `Error.nested` is present, it is preferred, preserving existing behavior.

## Type of change

## Related issues

Fixes https://github.com/rollbar/rollbar.js/issues/1010

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/test/browser.transforms.test.js
+++ b/test/browser.transforms.test.js
@@ -30,6 +30,10 @@ function itemFromArgs(args) {
   return item;
 }
 
+function chromeMajorVersion() {
+  return parseInt(navigator.userAgent.match(/Chrome\/([0-9]+)\./)[1]);
+}
+
 describe('handleDomException', function() {
   it('should do nothing if not a DOMException', function(done) {
     var err = new Error('test');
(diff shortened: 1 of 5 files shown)


test/browser.transforms.test.js
```diff
@@ -452,6 +456,45 @@ describe('addBody', function() {
       });
     });
   });
+  describe('with error cause', function() {
+    // Error cause was introduced in Chrome 93.
+    if (chromeMajorVersion() < 93) return;
+
+    it('should create trace_chain', function(done) {
+      var causeErr = new Error('cause error');
```
ijsnow, line 464: I know all old tests use the old `var` syntax but is there a reason to continue using them as opposed to `const`/`let`?

ijsnow (approved):
Looks good to me! Nifty that its just a map of existing rollbar functionality.

test/browser.transforms.test.js
waltjones: Just consistency. I'd rather not have a mix until we're actually switching over to `const`/`let`.

PR #1017, opened 2022-04-11
Refactor to use current (3.x) async package

waltjones:
## Description of the change

The async package dependency needed to be updated from v1.x to 3.x to pick up a security update. There have been interface changes since then, so a small refactor was also needed to work with the current async interface.

The specific changes that affect rollbar.js happened in async v2.0.0:
> filter, reject, some, every, detect and their families like {METHOD}Series and {METHOD}Limit now expect an error as the first callback argument, rather than just a simple boolean.

https://github.com/caolan/async/blob/master/CHANGELOG.md#breaking-changes-1

This PR updates the calling convention for callbacks, and removes usage of `fs.exists` which is is both deprecated and incompatible with the current async interface.

## Type of change

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -672,9 +672,9 @@
       "dev": true
     },
     "async": {
-      "version": "1.2.1",
-      "resolved": "https://registry.npmjs.org/async/-/async-1.2.1.tgz",
-      "integrity": "sha1-pIFqF81f9RbfosdpikUzabl5DeA="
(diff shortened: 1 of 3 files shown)


cyrusradfar (approved):
Thank you Walt.

waltjones:
Closed here and merged as https://github.com/rollbar/rollbar.js/pull/1018

Issue #1036, opened 2022-07-27
Wrapped Rollbar Node SDK running into timeout issues


phil-losty-rb:
Rollbar API timeouts are a problem for us. We've wrapped the Node SDK in a generous 200ms timeout and even that frequently is not enough. We were forced to add that timeout wrapper because Rollbar is sometimes so slow that minor warnings in our system become timeouts for our customers.

PR #1037, opened 2022-08-01
Perform occurrence POST write async

waltjones:
## Description of the change

There are concerns that when sending an occurrence, the write of the POST request payload can be blocked, blocking the caller. This is specifically about the request write not completing, not about any delay or timeout related to the response, which is already async.

While the http response completes async, the request is performed on the caller's thread. There are concerns that the caller could be blocked in some cases.

This PR uses a zero setTimeout to get off the caller's tick before making the request. This PR also updates all tests that relied on the request being synchronous. (Typically tests that inspect the request.)

Notes:
This PR adheres to this project's es5 syntax.

The execution is kept synchronous until the payload is fully constructed. Making these paths async would cause an undesired change in behavior where objects such as custom data would change value before being evaluated for the payload.

If the subsequent network write blocks the event loop, that may still lead to a perf impact, but that is outside the scope of this PR. 

## Type of change

## Related issues

Fixes: https://app.shortcut.com/rollbar/story/112245
Fixes: https://github.com/rollbar/rollbar.js/issues/1036

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/test/browser.core.test.js
+++ b/test/browser.core.test.js
@@ -43,23 +43,26 @@ describe('options.captureUncaught', function() {
 
     var element = document.getElementById('throw-error');
     element.click();
-    server.respond();
 
-    var body = JSON.parse(server.requests[0].requestBody);
+    setTimeout(function() {
+      server.respond();
 
-    expect(body.access_token).to.eql('POST_CLIENT_ITEM_TOKEN');
-    expect(body.data.body.trace.exception.message).to.eql('test error');
-    expect(body.data.notifier.diagnostic.raw_error.message).to.eql('test error');
-    expect(body.data.notifier.diagnostic.is_uncaught).to.eql(true);
+      var body = JSON.parse(server.requests[0].requestBody);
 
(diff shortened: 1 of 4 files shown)


test/browser.core.test.js
```diff
@@ -43,23 +43,26 @@ describe('options.captureUncaught', function() {
 
     var element = document.getElementById('throw-error');
     element.click();
-    server.respond();
 
-    var body = JSON.parse(server.requests[0].requestBody);
+    setTimeout(function() {
```
mudetroit, line 47: __*Comment*__
This strategy should work find 99% of the time, just watch out for flakiness. I don't think that there is a guarantee on execution order based on the length of the timeout

mudetroit (approved):
Yeah what you are doing here makes sense to me. As per my note, watch out for any flakiness on those setTimeouts with the tests, but that may take time to expose itself.

test/browser.core.test.js
waltjones: True, there's no guarantee.

PR #1055, opened 2022-11-29
Fix response body capture for fetch telemetry

waltjones:
### Background
When enabled via the Rollbar config, network telemetry can capture the response body for fetch requests. When this was originally enabled, it worked in browser, but was finicky depending on the version of headless chrome used in the tests. Recently, the previously working test started failing.

### Root cause
When the response body is a readable stream, the Rollbar handler must clone the response before reading the stream. The stream is locked once read, and cloning the stream allows safely reading it without reading the original. The code path already did this, but there's another requirement for `response.clone()`. It must be the first access to the response, or it will throw a `TypeError`.

The test started failing because headless chrome no longer accepted the readable stream used in the mock, and fails by throwing `TypeError`, mimicking the error described above. (Normally the browser would provide the readable stream, and will accept its own stream object.)  

### The fix
This PR changes the test mock to use a string for the response body, rather than a `ReadableStream` as was being used before. Previously, `ReadableStream` was the only way to get headless chrome to enforce the locking of the response body that we want to test. Currently, this behavior can be tested using a string in the mock, and both the first access requirement and single reader requirement are enforced and can be tested.

## Type of change

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/browser/telemetry.js
+++ b/src/browser/telemetry.js
@@ -383,7 +383,12 @@ Instrumenter.prototype.instrumentNetwork = function() {
         if (self.trackHttpErrors()) {
           metadata.stack = (new Error()).stack;
         }
-        return orig.apply(this, args).then(function (resp) {
+
+        var promise = orig.apply(this, args);
+
+        // Start our handler before returning the promise. This allows resp.clone()
+        // to execute before other handlers touch the response.
+        promise.then(function (resp) {
           metadata.end_time_ms = _.now();
           metadata.status_code = resp.status;
           metadata.response_content_type = resp.headers.get('Content-Type');
@@ -395,6 +400,7 @@ Instrumenter.prototype.instrumentNetwork = function() {
           if (self.autoInstrument.networkResponseBody) {
             if (typeof resp.text === 'function') { // Response.text() is not implemented on some platforms
               // The response must be cloned to prevent reading (and locking) the original stream.
+              // This must be done before other handlers touch the response.
               body = resp.clone().text(); //returns a Promise
             }
           }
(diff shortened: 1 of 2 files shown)


src/browser/telemetry.js
```diff
@@ -383,7 +383,12 @@ Instrumenter.prototype.instrumentNetwork = function() {
         if (self.trackHttpErrors()) {
           metadata.stack = (new Error()).stack;
         }
-        return orig.apply(this, args).then(function (resp) {
+
+        var promise = orig.apply(this, args);
+
+        // Start our handler before returning the promise. This allows resp.clone()
+        // to execute before other handlers touch the response.
+        promise.then(function (resp) {
```
mudetroit, line 391: __*Comment*__

Minor suggestion here. I don't think there are strong guarantees on the execution order of handlers to a single promise. Meaning in this case I don't think you can guarantee handlerA runs before handlerB:

```
function handlerA(response) { return response }
function handlerB(response) { return response }

var promise = new Promise(){...}

promise.then(handlerA)
promise.then(handlerB)
```

If your first handler returns the same response as the original (which yours does) you can simply return the promise returned by then to guarantee order

```
function handlerA(response) { return response }
function handlerB(response) { return response }

var promise = new Promise(){...}

var newPromise = promise.then(handlerA)
newPromise.then(handlerB)
```

mudetroit (commented):
Your logic all makes sense to me here, and great job with the explanation. I made one suggestion that may avoid what is probably a pretty rare edge case but perhaps something to consider.

src/browser/telemetry.js
waltjones: I agree this works based on existing browser behavior, not guaranteed. Great suggestion, I'll update the PR.

mudetroit (approved):

PR #1087, opened 2023-01-05
Update package.json

PraveenAnaparthi:
bump request-ip from 2.x to 3.3.0

## Description of the change
bump request-ip from 2.x to 3.3.0,
2. x version has a dependency [is_js](https://www.npmjs.com/package/is_js)  and it has a Vulnerability. I guess there are no breaking changes between 2. x and 3. x
https://ossindex.sonatype.org/component/pkg:npm/is_js@0.9.0?utm_source=dependency-check&utm_medium=integration&utm_content=7.4.3
For the past 6 years, there is no update for the is_js library. 

## Type of change
- [ ] Maintenance

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "error-stack-parser": "^2.0.4",
     "json-stringify-safe": "~5.0.0",
     "lru-cache": "~2.2.1",
-    "request-ip": "~2.0.1",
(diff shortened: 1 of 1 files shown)


waltjones (approved):
@PraveenAnaparthi I agree, there don't appear to be breaking changes in the changelog. Thank you for the PR!

PR #1102, opened 2023-04-20
fix: exit server after reconfiguring

alecgibson:
According to [the docs][1], `exitOnUncaughtException` should:

> Exit the process after handling an uncaught exception. Requires
> `captureUncaught` to also be set.

So with this config:

````js
var rollbar = new Rollbar({
  captureUncaught: true,
  exitOnUncaughtException: true,
  // ...
});
```

...an uncaught exception should exit the process.

This works correctly, but *only* if `rollbar.configure()` is *never* called again, because:

 1. `exitOnUncaughtException` [was deleted][2] from our options
 2. Calling `configure()` calls [`setupUnhandledCapture()`][3]...
 3. ...which calls [`handleUncaughtExceptions()`][4]...
 4. ...which now [tries to access][5] `options.exitOnUncaughtException`
 5. But this option was deleted in Step 1, so when we [replace][6] our `uncaughtException` handler, we now have `exitOnUncaught = false`

This change:

 - no longer deletes `exitOnUncaughtException` from options (it's unclear why this is the only option that is deleted)
 - directly accesses `options.exitOnUncaughtException` in the handler, just like `options.captureUncaught` is accessed directly

[1]: https://docs.rollbar.com/docs/rollbarjs-configuration-reference#context-1
[2]: https://github.com/rollbar/rollbar.js/blob/e964ecbdb16a4d2b8b5feaa8b70a2ec327648ed0/src/server/rollbar.js#L587
[3]: https://github.com/rollbar/rollbar.js/blob/e964ecbdb16a4d2b8b5feaa8b70a2ec327648ed0/src/server/rollbar.js#L118
[4]: https://github.com/rollbar/rollbar.js/blob/e964ecbdb16a4d2b8b5feaa8b70a2ec327648ed0/src/server/rollbar.js#L578
[5]: https://github.com/rollbar/rollbar.js/blob/e964ecbdb16a4d2b8b5feaa8b70a2ec327648ed0/src/server/rollbar.js#L586
[6]: https://github.com/rollbar/rollbar.js/blob/e964ecbdb16a4d2b8b5feaa8b70a2ec327648ed0/src/server/rollbar.js#L589

## Description of the change

> Please include a summary of the change and which issues are fixed.
> Please also include relevant motivation and context.

## Type of change

- [ ] Maintenance
- [ ] New release

## Related issues

> Shortcut stories and GitHub issues (delete irrelevant)

- Fix [SC-]
- Fix #1

## Checklists

### Development

- [ ] Lint rules pass locally

### Code review


alecgibson:
@waltjones any chance of please getting this reviewed? We've still got an ugly patch in our code to work around this issue.

PR #1145, opened 2024-10-03
Fix Vue 3 loop

### LINKED ISSUE #1126: Rollbar in Vue app causes freeze / 100% CPU usage (inf. recursion)
I have a color picker in my site which shows up on clicking an icon. The site is built using **Vue** and is integrated with **Rollbar**. Though rollbar is working fine and is able to log errors to the dashboard. But few days back I encountered a weird problem with rollbar, when I mistakenly missed to add the input tag for color picker.

image

**What is expected?**
Ideally there should be an error and it should be logged in rollbar. 

**What is happening?**
There's and infinite log of warning from rollbar in the console and the CPU utilisation peaks up to 100% and the site freezes.

image
image
image

**Removing Rollbar** -
Removing rollbar integration doesn't freezes the site and logs the following warning and error in console -

image

image

**Is there anything wrong with Rollbar internals? or is there a way I can tackle this problem? Thanks, Ahead!**

b3bb0:
Fix loop in vue 3

## Description of the change

> in **Vue 3**: when rollbar is triggered by a throw error from inside a helper or the router
> (I suspect also some other odd placed) including the vm variable can generate an infinte loop
> The (proxy) data included in vm are not really needed for debug purposes

> Allso added new. tiny feature to fire on warnings from vue

## Type of change

- [ ] Maintenance
- [ ] New release

## Related issues

- Fix #1126

### Development

- [x] Lint rules pass locally
- [x ] All tests related to the changed code pass in development

### Code review

### DIFF
--- a/examples/vuejs3/src/rollbar.js
+++ b/examples/vuejs3/src/rollbar.js
@@ -7,11 +7,23 @@ const rollbar = new Rollbar(config);
 export default {
   install(app) {
     app.config.errorHandler = (error, vm, info) => {
-      rollbar.error(error, { vueComponent: vm, info });
-      if (app.config.devtools) {
-        console.error(error);
-      }
-    };
+      
+      // In case the error is from the router or am helper
+      // calling vm could generate a loop and freeze the browser
+      // rollbar.error(error, { vueComponent: vm, info });
+      
+      rollbar.error(error, { info });
+      
+      if (app.config.devtools) console.error(error);
+    }
+
+    
+    app.config.warningHandler = (error, vm, info) => {
+      rollbar.warning(error, { info });
+      
+      if (app.config.devtools) console.log('just a warning, but!,error)
+    }
+    
     app.provide('rollbar', rollbar);
   },
 };


examples/vuejs3/src/rollbar.js
```diff
@@ -7,11 +7,23 @@ const rollbar = new Rollbar(config);
...
+      // rollbar.error(error, { vueComponent: vm, info });
+      
+      rollbar.error(error, { info });
+      
+      if (app.config.devtools) console.error(error);
+    }
+
+    
+    app.config.warningHandler = (error, vm, info) => {
+      rollbar.warning(error, { info });
```
itsalaidbacklife, line 22: I think a more appropriate example is to grab the most relevant and common data off the component, with coalesced fallbacks to empty data.

I also think it's best to stick with the existing format of multiline conditionals. It's a style choice, but the more common one, and since that's how the example was previously styled, I think we should stick to it for consistency across the docs
```suggestion
      // In case the error is from the router or a helper
      // calling vm could generate a loop and freeze the browser
      // Instead, we pick the relevant data off the vm instance
      
       const vueComponent = {
              name: vm.$options?.name ?? 'AnonymousComponent',
              props: vm.$props ?? {},
              data: vm.$data ?? {},
              attributes: vm.$attributes ?? {},
              route: vm.$route ?? {},
      };
      
      rollbar.error(error, { vueComponent, info });
      
       if (app.config.devtools) {
              console.error(error);
      }

    
    app.config.warningHandler = (error, vm, info) => {
      rollbar.warning(error, { vueComponent, info });
```

examples/vuejs3/src/rollbar.js
```diff
@@ -7,11 +7,23 @@ const rollbar = new Rollbar(config);
...
+      
+      if (app.config.devtools) console.log('just a warning, but!,error)
```
itsalaidbacklife, line 24: For consistency across the docs, I suggest we keep this to multiline format
```suggestion
      if (app.config.devtools) {
              console.log('just a warning, but!, error)
      }
```

itsalaidbacklife:
I agree that the docs should be changed. This is a fairly egregious oversight as the introductory docs for vue apps literally tell you to log in a way that will crash your application.

That said I think that this should be renamed and not necessarily considered a resolution to #1126 since it only updates the docs (which we should do), but ideally `rollbar` should update in such a way as to handle proxy vue 3 components so as not to crash the app if you tried to log a full component. I'm not sure if this is possible, but given what a footgun 'logging your component instance' is, it would be nice if some guard rails could be put in place besides just the example documentation

waltjones (approved):
@b3bb0 Thank you for the PR!

PR #1151, opened 2025-02-21
perf: make Zone check framework agnostic

arturovt:
In this commit, we prevent coupling the Zone check to Angular, as zone.js can be used with any framework. We always ensure that we get the root zone and run the callback within it.


waltjones (approved):
@arturovt Thank you for the PR!

PR #1165, opened 2025-04-07
Sessions and tracing

waltjones:
## Description of the change

Adds OTel compatible tracing support. This implementation aims to be lightweight and simple to use in browser applications that don't want to pull in all of the opentelelemetry-js packages.

* This PR adds trace, span, and context management.
* Exporter will com in a follow up PR.
* Tests will com in a follow up PR.

## Type of change

### DIFF
--- a/src/browser/core.js
+++ b/src/browser/core.js
@@ -16,17 +16,22 @@ var errorParser = require('../errorParser');
 function Rollbar(options, client) {
   this.options = _.handleOptions(defaultOptions, options, null, logger);
   this.options._configuredOptions = options;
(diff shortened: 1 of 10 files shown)


matux (approved):
This is the single best PR I've ever seen in my life.

Some notes:
- I'd love it if we could eventually get docstrings (maybe something ai could do)
- Would love to figure out if we can do async/await for the `with` functions

We went through this code with @waltjones and @shankj3 in a talk and it all looked pretty good, the implementation is clean and simplifies interacting with opentelemetry.

So, lovely.

PR #1166, opened 2025-04-10
Initial tests for tracing

waltjones:
## Description of the change

Adds initial tests for tracing. This PR also includes a couple of light refactors related to the tests.

## Type of change

- [x] Refactor
- [x] Test

### Development

- [x] Lint rules pass locally

### DIFF
--- a/src/tracing/span.js
+++ b/src/tracing/span.js
@@ -17,7 +17,7 @@ export class Span {
       kind: options.kind,
       spanContext: options.spanContext,
       parentSpanId: options.parentSpanId,
-      startTime: this.hrTimeNow(),
+      startTime: options.startTime || this.hrTimeNow(),
       endTime: [0, 0],
       status: {code: 0, message: ''},
       attributes: {'session.id': options.session.id},
@@ -52,13 +52,13 @@ export class Span {
     return this;
   }
 
-  addEvent(name, attributes) {
+  addEvent(name, attributes = {}, time) {
     if (this.span.ended) return this;
 
     this.span.events.push({
       name,
       attributes,
-      time: this.hrTimeNow(),
+      time: time || this.hrTimeNow(),
       droppedAttributesCount: 0,
     });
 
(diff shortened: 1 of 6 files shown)


src/tracing/span.js
```diff
@@ -52,13 +52,13 @@ export class Span {
     return this;
   }
 
-  addEvent(name, attributes) {
+  addEvent(name, attributes = {}, time) {
```
matux, line 55: How does this work in js?
```js
someTime = 1234
addEvent('name', someTime)
```
Does the value of `someTime` binds to `attributes` or `time`?

src/tracing/spanProcessor.js
```diff
@@ -1,16 +1,15 @@
...
 
   onStart(span, _parentContext) {
-    pendingSpans.set(span.spanContext.spanId, span);
+    this.pendingSpans.set(span.span.spanContext.spanId, span);
   }
 
   onEnd(span) {
     this.exporter.export([span.export()])
-    pendingSpans.delete(span.spanContext.spanId);
+    this.pendingSpans.delete(span.span.spanContext.spanId);
```
matux, line 13: There's a span in my span! What type is the `span` parameter?

src/tracing/span.js
msandula12: Your `someTime` argument would bind to `attributes`.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters):

> Parameters are still set left-to-right, overwriting default parameters even if there are later parameters without defaults.

waltjones: It's the second argument, so `attributes`.

src/tracing/spanProcessor.js
waltjones: It's a `Span` instance.

matux: Oh, got it. So one is the `Span`, and the other is the opentelemetry key/value pairs? Kind of like the "raw" description of the span.

I don't know if it's a JS thing, but I find it a bit weird that `Span` has a `span` instance var that's not a `Span`. Why not call it `rawSpan` or something more explicit?

src/tracing/span.js
matux: Talked offline, we're gonna do a param object refactor ala react to have labels in params.

src/tracing/spanProcessor.js
matux: Won't change or change later.

matux (approved):

PR #1183, opened 2025-05-09
Add trace attributes to item

waltjones:
## Description of the change

* Adds a general utility for adding attributes to an occurrence.
* Adds span attributes to the occurrence for session_id, span_id, and trace_id if there is an active span at the time of the occurrence.
* Adds an event to the current span for the occurrence.

## Type of change

### DIFF
--- a/src/rollbar.js
+++ b/src/rollbar.js
@@ -10,7 +10,7 @@ var _ = require('./utility');
  * @param api
  * @param logger
  */
-function Rollbar(options, api, logger, telemeter, platform) {
+function Rollbar(options, api, logger, telemeter, tracing, platform) {
   this.options = _.merge(options);
   this.logger = logger;
   Rollbar.rateLimiter.configureGlobal(this.options);
@@ -18,6 +18,9 @@ function Rollbar(options, api, logger, telemeter, platform) {
   this.api = api;
   this.queue = new Queue(Rollbar.rateLimiter, api, logger, this.options);
 
+  this.tracing = tracing;
+
+  // Legacy OpenTracing support
   // This must happen before the Notifier is created
   var tracer = this.options.tracer || null;
   if (validateTracer(tracer)) {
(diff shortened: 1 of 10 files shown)


src/rollbar.js
```diff
@@ -164,6 +171,26 @@ Rollbar.prototype._log = function (defaultLevel, item) {
   }
 };
 
+Rollbar.prototype._addTracingAttributes = function (item) {
```
shankj3, line 174: so simple, nice

shankj3 (approved):

src/tracing/span.js
```diff
@@ -37,6 +37,14 @@ export class Span {
     return this.span.spanContext;
   }
 
+  spanId() {
```
matux, line 40:
```suggestion
  get spanId() {
```
Can be used like a property, since it's not doing any computation, I think it can be a minor improvement.

##### non-blocking

src/tracing/span.js
```diff
@@ -37,6 +37,14 @@ export class Span {
...
+    return this.span.spanContext.spanId;
+  }
+
+  traceId() {
```
matux, line 44:
```suggestion
  get traceId() {
```
Same as above.

src/tracing/tracing.js
```diff
@@ -23,6 +23,14 @@ export default class Tracing {
     }
   }
 
+  sessionId() {
```
matux, line 26:
```suggestion
  get sessionId() {
```

src/rollbar.js
matux: Agreed!

src/rollbar.js
```diff
@@ -164,6 +171,26 @@ Rollbar.prototype._log = function (defaultLevel, item) {
...
+  if (this.tracing) {
+    const span = this.tracing.getSpan();
+    if (!span) {
+      return;
+    }
```
matux, line 179: Could take advantage of this.tracing being null and just do:
```suggestion
  const span = this.tracing?.getSpan();
  if (!span) return;
```
Not sure though.

##### non-blocking

matux (approved):
Looks great!!!

Some tiny weeny non-blocking language comments that I recently learned about JS 😁

src/tracing/span.js
waltjones: I agree, this is better.

PR #1184, opened 2025-05-09
Add path aliases for cleaner imports

matux:
## Description of the change

I was getting tired of the relative subdirs when doing imports and checked if there's an idiomatic way of improving how we can refer to different paths without crawling for directories. Turns out there's, path aliases.

Added some configuration and updated the imports and it seems to be working just fine. Although I don't know if this could generate other issues for SDK users.

## Type of change

- [x] Maintenance
- [ ] New release

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -82,6 +82,17 @@ module.exports = function (config) {
           },
         ],
       },
+      resolve: {
+        alias: {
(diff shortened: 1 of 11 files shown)


waltjones (changes requested):
This would require any app using absolute imports to support path aliases.

PR #1186, opened 2025-05-10
Extracted hrtime utilities in tracing to its own module

matux:
## Description of the change

This PR introduces a dedicated utility module, `hrtime.js` and a new way to interface with it:
```js
import hrtime from '@tracing/hrtime.js';

hrtime.fromMillis(1000);
hrtime.toMillis([0, 1000]);
hrtime.add([0, 0], [0, 1000]);
hrtime.now();
hrtime.isHrTime([0, 1000]);
```

The reason is that I've been needing to use this in multiple modules and I don't see why its access should be dependent upon an instance of `Span`.

- Renames `sum` to `add` to follow OTEL convention @ https://github.com/open-telemetry/opentelemetry-js/blob/0c21db4621fd981200cc731d676d1a87162ee0d9/packages/opentelemetry-core/src/common/time.ts#L173-L183

## Type of change

- [x] Maintenance
- [ ] New release

### DIFF
--- a/src/browser/replay/recorder.js
+++ b/src/browser/replay/recorder.js
@@ -1,6 +1,8 @@
 import { record as rrwebRecordFn } from '@rrweb/record';
 import { EventType } from '@rrweb/types';
 
+import hrtime from '../../tracing/hrtime.js';
+
 export default class Recorder {
   #tracing;
   #options;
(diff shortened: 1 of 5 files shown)


waltjones (approved):
Great idea.

PR #1190, opened 2025-05-13
Update postSpans

waltjones:
## Description of the change

This PR updates `postSpans` to send the expected request format, and updates the options management to allow the correct API url path to be used. (The transport options logic is such that `options.endpoint` will always override, causing the item endpoint to be used. This PR introduces a new config option at `options.tracing.endpoint` and ensures that it is used instead.)

This PR also includes related improvements.
* Remove access token from the payload for both occurrences and traces.
* Remove jsonBackup, which is no longer needed.
* Updates tests to match these changes.

## Type of change

- [x] Maintenance

### DIFF
--- a/src/tracing/defaults.js
+++ b/src/tracing/defaults.js
@@ -1,0 +1,7 @@
+/**
+ * Default tracing options
+ */
+export default {
+  enabled: false,
+  endpoint: 'api.rollbar.com/api/1/session/',
+}
--- a/src/api.js
+++ b/src/api.js
@@ -10,7 +10,7 @@ var defaultOptions = {
   port: 443,
 };
 
-var sessionDefaultOptions = {
+var OTLPDefaultOptions = {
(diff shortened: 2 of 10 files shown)


src/tracing/defaults.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Default tracing options
+ */
+export default {
+  enabled: false,
+  endpoint: 'api.rollbar.com/api/1/session/',
```
matux, line 6: Wonder if we should just define `/api/1/session/` and leave the host out. For example if we want to test this in our localhost? I assume the host is set in some other global config place?

matux (approved):
Nice! One question re the tracing endpoint config having the host. Non-blocking.

src/tracing/defaults.js
waltjones: Let's simplify transport options in general in a follow up PR.

PR #1194, opened 2025-05-13
Update resource and (temporary) scope

waltjones:
## Description of the change

Fixes the options for resource attributes and scope. The scope is temporary, and will likely be updated to use `telemetry.sdk.*` attributes, while making the scope name and version available to the app.

## Type of change


matux (approved):
Fantastic! Now I got what you meant. Perfect!

PR #1195, opened 2025-05-14
Add item.uuid to replay span

waltjones:
## Description of the change

Adds the `occurrence.uuid` attribute to the replay span.

## Type of change


src/browser/replay/recorder.js
```diff
@@ -75,6 +75,10 @@ export default class Recorder {
 
     recordingSpan.setAttribute('rollbar.replay.id', replayId);
 
+    if (occurrenceUuid) {
+      recordingSpan.setAttribute('occurrence.uuid', occurrenceUuid);
```
shankj3, line 79: Since this is an attribute on the span itself and not an event attribute for our custom events, I think it makes sense to prefix. Sorry for the runaround there, i'll also update mine

shankj3 (approved):

PR #1196, opened 2025-05-14
Fix resource attributes

waltjones:
## Description of the change

* Wrap resource attributes, so the exporter can consume them.
* Remove the `options.resource` override in the exporter.

## Type of change

### DIFF
--- a/src/tracing/exporter.js
+++ b/src/tracing/exporter.js
@@ -34,7 +34,7 @@ export class SpanExporter {
       return { resourceSpans: [] };
     }
 
-    const resource = options.resource || (spans[0] && spans[0].resource) || {};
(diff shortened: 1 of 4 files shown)


src/tracing/exporter.js
```diff
@@ -34,7 +34,7 @@ export class SpanExporter {
       return { resourceSpans: [] };
     }
 
-    const resource = options.resource || (spans[0] && spans[0].resource) || {};
+    const resource = (spans[0] && spans[0].resource) || {};
```
matux, line 37: Could you update the jsdoc, too? In fact, I think we could remove the options param altogether.

matux (changes requested):
This is perfect, we should update the jsdoc, tho.

PR #1199, opened 2025-05-22
Initial timeline events (otel telemetry events)

waltjones:
## Description of the change

Initial support for 'timeline events', sending telemetry events in an otel span.

The 'rollbar-telemetry' span is ended and a new one started whenever an occurrence is sent. This ensures a span is available to send when a replay is sent.

Note:
I considered adding the events in `Telemeter.prototype.capture`, but it seems there is enough differentiation in the event types that it's better to do it in the path for each event type.

## Type of change

### DIFF
--- a/src/telemetry.js
+++ b/src/telemetry.js
@@ -1,12 +1,15 @@
 var _ = require('./utility');
+const hrtime = require('./tracing/hrtime').default;
 
-var MAX_EVENTS = 100;
+const MAX_EVENTS = 100;
 
-function Telemeter(options) {
+function Telemeter(options, tracing) {
   this.queue = [];
   this.options = _.merge(options);
   var maxTelemetryEvents = this.options.maxTelemetryEvents || MAX_EVENTS;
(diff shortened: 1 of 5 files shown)


src/telemetry.js
```diff
@@ -87,12 +90,25 @@ Telemeter.prototype.captureError = function (
   rollbarUUID,
   timestamp,
 ) {
-  var metadata = {
-    message: err.message || String(err),
-  };
+  const message = err.message || String(err);
+  var metadata = {message};
```
matux, line 94: I think metadata can be `const`. Const prevents reassignment, not mutation.

matux (approved):
I find no fault in here, just a tiny non-blocking comment.

It'd be cool to have an e2e test that combines both rrweb events + telemetry.

src/telemetry.js
waltjones: Right. That was an existing line I updated and overlooked the `var`.

PR #1200, opened 2025-05-27
Add `rollbar.replay.id` attribute to telemetry span

waltjones:
## Description of the change

Adds the replay_id to the telemetry span. This may not be a long term solution, but is what the back end currently expects.

In this PR:
* Allow the replayMap to optionally accept an existing replayId.
* Add a `genId` convenience method to Tracing.
* Add `rollbar.replay.id` to the telemetry span.

## Type of change


matux (approved):
Looks good

PR #1207, opened 2025-06-04
Migrate cjs to esm: utility, core, general, platform and browser tests

matux:
## Description of the change

This PR migrates all test files from CommonJS to ES modules—except for Vows-related server tests—by replacing require() calls with import statements, appending .js extensions, and updating async patterns where needed.

## Type of change

- [x] Maintenance
- [ ] New release

## Related issues

[SDK-489/partial-test-migration-to-es-modules](https://linear.app/rollbar-inc/issue/SDK-489/partial-test-migration-to-es-modules)

### DIFF
--- a/test/browser.transforms.test.js
+++ b/test/browser.transforms.test.js
@@ -3,8 +3,8 @@
 /* globals it */
 /* globals sinon */
 
-var Rollbar = require('../src/browser/rollbar');
-var t = require('../src/browser/transforms');
+import Rollbar from '../src/browser/rollbar.js';
+import t from '../src/browser/transforms.js';
 
 function TestClientGen() {
   var TestClient = function () {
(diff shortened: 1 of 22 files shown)


test/browser.transforms.test.js
```diff
@@ -547,16 +547,23 @@ describe('scrubPayload', function () {
...
-      done(e);
+    const scrubModule = await import('../src/scrub.js');
+    const scrub = scrubModule.default;
+
+    const scrubberFn = t.addScrubber(scrub);
+    const result = await new Promise((resolve, reject) => {
+      scrubberFn(payload, options, (err, result) => {
+        if (err) reject(err);
+        else resolve(result);
+      });
```
matux, line 558: I'm not sure if this is the best way of doing this, in node you can use `promisify`, but I don't know if there's an equivalent available in this environment.

waltjones: https://github.com/mikehall314/es6-promisify as a dev dependency would be ok, but I think this is fine as is.

waltjones (approved):

PR #1217, opened 2025-06-09
Send/discard replay based on item response headers

waltjones:
## Description of the change

Uses `Rollbar-Replay-Enabled` and `Rollbar-Replay-RateLimit-Remaining` response headers from the post item response to determine whether to send the associated replay. The `Rollbar-Replay-RateLimit-Reset` header is ignored for now.

## Type of change

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/queue.js
+++ b/src/queue.js
@@ -322,7 +322,6 @@ Queue.prototype._maybeCallWait = function () {
  * @private
  */
 Queue.prototype._handleReplayResponse = async function (replayId, response, headers) {
-  console.log('Queue._handleReplayResponse:', replayId, response, headers);
   if (!this.replayMap) {
     console.warn('Queue._handleReplayResponse: ReplayMap not available');
     return false;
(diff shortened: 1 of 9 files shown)


src/queue.js
```diff
@@ -321,7 +321,8 @@ Queue.prototype._maybeCallWait = function () {
  *                             false if replay was discarded or an error occurred
  * @private
  */
-Queue.prototype._handleReplayResponse = async function (replayId, response) {
+Queue.prototype._handleReplayResponse = async function (replayId, response, headers) {
+  console.log('Queue._handleReplayResponse:', replayId, response, headers);
```
matux: Do we want this?

matux (approved):
There's a `console.log` we may want to remove.

src/queue.js
waltjones: No good catch.

PR #1227, opened 2025-06-25
Only extract replay headers in xhr and fetch when appropriate

matux:
## Description of the change

Only get replay related headers when the response is to an item request.

## Type of change

- [ ] Maintenance
- [ ] New release

### DIFF
--- a/src/browser/transport/xhr.js
+++ b/src/browser/transport/xhr.js
@@ -31,17 +31,22 @@ function makeXhrRequest(
 
             var parseResponse = _.jsonParse(request.responseText);
             if (_isSuccess(request)) {
-              const headers = {
-                'Rollbar-Replay-Enabled': request.getResponseHeader(
-                  'Rollbar-Replay-Enabled'
-                ),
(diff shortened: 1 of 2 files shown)


src/browser/transport/xhr.js
```diff
@@ -157,14 +162,12 @@ function _createXMLHTTPObject() {
   var i;
   var numFactories = factories.length;
   for (i = 0; i < numFactories; i++) {
-    /* eslint-disable no-empty */
```
mudetroit: 🤨

mudetroit (commented):
Primary change looks fine. What is with the unrelated eslint change?

src/browser/transport/xhr.js
matux: Yeah, I presume this was code auto-removing an unused eslint directive, ie. no "no-empty" are ever produced. I'll give it a closer look domani.

matux: I restored these.

PR #1235, opened 2025-07-11
Send custom 'end' event

waltjones:
## Description of the change

Helps the application correctly align playback by adding a noop event to the end of the recording.

## Type of change

### DIFF
--- a/src/browser/replay/recorder.js
+++ b/src/browser/replay/recorder.js
@@ -116,6 +116,8 @@ export default class Recorder {
       );
     }
 
+    this._addEndEvent(recordingSpan, replayId);
+
     recordingSpan.end();
 
     return tracing.exporter.toPayload();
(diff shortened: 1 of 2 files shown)


src/browser/replay/recorder.js
```diff
@@ -191,4 +193,21 @@ export default class Recorder {
...
+   * Helps the application correctly align playback by adding a noop event
+   * to the end of the recording.
+   **/
+  _addEndEvent(recordingSpan, replayId) {
+    recordingSpan.addEvent(
+
+      'rrweb-replay-events',
+      {
+        eventType: 5,
+        json: JSON.stringify({tag: "replay.end", payload: {}}),
```
shankj3, line 207: This won't help us in the case of someone closing a tab, right? 
Ie previous recording events have already been sent for this replay, then one more is going to be sent in the near future but tab is closed

shankj3 (approved):

src/browser/replay/recorder.js
waltjones: This span either gets sent in whole, or not at all. The purpose of adding this event is so that the player extends its playback until the triggering occurrence occurs. (It's almost like giving the occurrence an rrweb event.)

PR #1238, opened 2025-07-14
Add type definitions for session replay related types

matux:
## Description of the change

This PR adds missing type definitions for the new Session Replay related classes, `Recorder` and `Tracing`. Following the same pattern as we currently have.

This is necessary for TS projects in order to be able to pass `recorder` and `tracing` options to `Rollbar` at the point of instantiation.

## Type of change

- [ ] Maintenance
- [ ] New release

## Related issues

[SDK-517](https://linear.app/rollbar-inc/issue/SDK-517/add-type-definitions-for-session-replay-types)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -107,6 +107,7 @@ declare namespace Rollbar {
     ) => void;
     overwriteScrubFields?: boolean;
     payload?: Payload;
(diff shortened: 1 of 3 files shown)


brianr (approved):
👍

PR #1239, opened 2025-07-15
Initial (minimal) support for replay triggers config

waltjones:
## Description of the change

Replaces `triggerOptions` with a minimal initial implementation of the `triggers` array, which can support multiple types of triggers with different configurations.

The default if no `triggers` key is present is, as before, an occurrence trigger on `error` and `critical` level.

This initial implementation allows users of the next alpha release to use the new config to modify or disable the occurrence trigger.

## Type of change

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -238,11 +238,10 @@ declare namespace Rollbar {
     enabled?: boolean;
     autoStart?: boolean;
     maxSeconds?: number;
-    triggerOptions?: {
-      item?: {
-        levels?: Level[];
-      };
-    };
+    triggers?: {
+      type: string;
+      level?: string[];
+    }[];
     debug?: {
       logErrors?: boolean;
       logEmits?: boolean;
--- a/index.d.ts
+++ b/index.d.ts
@@ -240,7 +240,7 @@ declare namespace Rollbar {
     maxSeconds?: number;
     triggers?: {
       type: string;
-      level?: string[];
+      level?: Level[];
     }[];
     debug?: {
       logErrors?: boolean;
--- a/src/rollbar.js
+++ b/src/rollbar.js
@@ -200,8 +200,9 @@ Rollbar.prototype._addTracingAttributes = function (item, replayId) {
 };
 
 Rollbar.prototype._replayIdIfTriggered = function (item) {
-  const levels = this.options.recorder?.triggerOptions?.item?.levels || [];
-  if (levels.includes(item.level)) {
+  const triggers = _.getReplayTriggersForType(this.options.recorder?.triggers, 'occurrence');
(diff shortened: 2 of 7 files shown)


src/utility.js
```diff
@@ -663,6 +663,19 @@ function getItemAttribute(itemData, key) {
...
+
+  const filteredTriggers = [];
+  for (const t of triggers) {
+    if (t.type === triggerType) {
+      filteredTriggers.push(t);
+    }
+  }
+
+  return filteredTriggers;
+}
```
matux, line 677: Does this need to be a function?
```js
const triggers = this.options.recorder?.triggers || [];
const occurrenceTriggers = triggers.filter(t => t.type === 'occurrence');
```
Seems pretty straightforward/self-documenting at the point of usage.

index.d.ts
```diff
@@ -238,11 +238,10 @@ declare namespace Rollbar {
...
     autoStart?: boolean;
     maxSeconds?: number;
-    triggerOptions?: {
-      item?: {
-        levels?: Level[];
-      };
-    };
+    triggers?: {
+      type: string;
+      level?: string[];
```
matux: We have a type for this `Level[]` that's defined as:
```ts
export type Level = 'debug' | 'info' | 'warning' | 'error' | 'critical';
```

src/rollbar.js
```diff
@@ -260,6 +261,15 @@ Rollbar.prototype._addTracingInfo = function (item) {
...
 };
 
+Rollbar.prototype._getEnabledTrigger = function (triggers, level) {
+  for (const t of triggers) {
+    if (t.level?.includes(level)) {
+      return t;
+    }
+  }
+  return null;
+};
```
matux, line 271: Does this need to be a function? Seems pretty straight-forward/self-documenting at the point of use:
```ts
const enabledTrigger = this._getEnabledTrigger(triggers, item.level);
```
Becomes
```ts
const enabledTrigger = triggers.find(t => t.level?.includes(level)) || null;
```
The code literally says "find the trigger with the levels that includes some `level`."

matux (commented):
Some suggestions, nothing really blocking, except I think we should use the `Level[]` type instead of `string` for `level` in `index.d.ts`.

src/utility.js
waltjones: It will be called from more that one place and will likely do more work than it does now.

src/rollbar.js
waltjones: There will be additional filters.

index.d.ts
waltjones: Updated

matux (approved):

PR #1246, opened 2025-07-18
Reorganize defaults

matux:
## Description of the change

This PR eliminates runtime `package.json` imports from the server and react native platforms, replacing them with proper ES module defaults. This fixes compatibility issues with modern build tools like Angular SSR, which cannot resolve `package.json` imports in certain contexts.

The Angular example now builds successfully without any changes to the example project itself.

The `DefinePlugin` webpack dependency has been removed together with all traces of the old way of loading defaults.

## Type of change

- [x] Maintenance
- [ ] New release

## Related issues

[SDK-518/make-sure-all-sdk-examples-work-properly](https://linear.app/rollbar-inc/issue/SDK-518/make-sure-all-sdk-examples-work-properly)


src/browser/plugins/jquery.js
matux, line 1: I'm not sure if this is right. Is this file meant to be standalone? Would the import work in this case? Would webpack do the smart thing and bundle the actual constant instead of leaving the import? Actually haven't checked this...

Should tests fail if this didn't work?

waltjones: Webpack writes a file to /dist/plugins/. You can `npm run build` and see if it generates the expected output.

Here is the usage doc. https://docs.rollbar.com/docs/rollbarjs-plugins

matux: Yep, it does the right thing!
```js
t.configure({payload:{notifier:{plugins:{jquery:{version:"0.0.8"}}}}});
```

PR #1256, opened 2025-07-22
Removed `json2` related files from vendor

matux:
## Description of the change

The eval-dependent `json2` vendor scripts aren't in use anymore, so we're removing them.

Additionally, `vendor/JSON-js/json_parse_state.js` was embedded into `vendor/JSON-js/json3.js`:
- Lines 64-406 of json_parse_state.js (the standalone state machine parser)
- Lines 418-759 of json3.js (the inline state machine parser)
They contain the same state machine implementation - the same variables, the same logic, the same structure. It appears that the code from json_parse_state.js was copied directly into `json3.js` rather than being imported.

The vendor's `README` has also been updated to reflect the current state of affairs.

Relevant Slack thread with archæological investigation: https://rollbar.slack.com/archives/C03C7UU8S/p1753176924757479

## Type of change

- [x] Maintenance
- [ ] New release


matux:
> Copilot wasn't able to review any files in this pull request.

Good job, buddy!

waltjones (approved):

PR #1265, opened 2025-07-30
Replay predicates, with probability sampling

waltjones:
## Description of the change

This PR implements replay predicates based on replay trigger configuration. In Rollbar.js, predicates decide whether a payload should be accepted or rejected.

The sampling algorithm is based on OTel trace id probability sampling, using the replay id instead of the trace id.
* https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/
* https://opentelemetry.io/docs/specs/otel/trace/tracestate-handling/

Notes: 
* String compare is more performant than conversion to float, assuming the `th` calculation will be moved to the trigger configuration. This allows `toString` to be called once, rather than `parseInt` to be called on each replay.
* `th` calculation will be moved in a follow up PR focused on config options.
 

## Type of change


src/browser/replay/replayPredicates.js
```diff
@@ -0,0 +1,89 @@
...
+   * @param {Object} trigger - The trigger object containing the sampling ratio.
+   * @returns {boolean} - True if the trigger is sampled, false otherwise.
+   */
+  isSampled(trigger) {
+    const ratio = trigger.samplingRatio || 1;
+
+    if (ratio == 1) {
+      return true;
+    }
+    const rv = this.context.replayId.slice(-7);
```
matux: Shouldn't this be 14?

[Sampling prob is 56 bytes/7 bytes.](https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/#sampling-probability) So I'm guessing that's where the 7 comes from, right? But isn't the `replayId` hex repr? So 14 hex chars.

Or did I forget how we're representing replay ids already? 😂

src/rollbar.js
```diff
@@ -200,10 +201,14 @@ Rollbar.prototype._addTracingAttributes = function (item, replayId) {
 };
 
 Rollbar.prototype._replayIdIfTriggered = function (item) {
-  const triggers = _.getReplayTriggersForType(this.options.recorder?.triggers, 'occurrence');
```
matux, line 203: I just checked and it seems we're not using this function anymore, maybe it could be removed from `src/utility.js`.

src/browser/replay/replayPredicates.js
```diff
@@ -0,0 +1,89 @@
...
+   * This allows `toString` to be called once, rather than `parseInt` to be called on
+   * each replay.
+   *
+   * @param {Object} trigger - The trigger object containing the sampling ratio.
+   * @returns {boolean} - True if the trigger is sampled, false otherwise.
+   */
+  isSampled(trigger) {
+    const ratio = trigger.samplingRatio || 1;
+
+    if (ratio == 1) {
```
matux, line 81: Just checking `==` is what we want and not `===`.

matux (commented):
Looks perfect, just one question about the sampling precision

src/browser/replay/replayPredicates.js
waltjones: We don't use ratio if this expression is true, so theres no harm in it being a different type. `==` allows it to be a BigInt for example.

src/browser/replay/replayPredicates.js
waltjones: Good catch.

matux (approved):

PR #1266, opened 2025-07-30
Webpack build orchestration

matux:
## Description of the change

This PR removes all responsibility from building the `dist` away from Grunt and we simply invoke webpack from the build script.

We also move the `update:snippets` script call to `postbuild`.

This PR contains the rebuilt `dist` from the latest `master` in order to ensure both grunt and the new method produce the exact same `dist`. Which it does.

## Type of change

- [x] Maintenance
- [ ] New release

## Related issues

[SDK-506/replace-grunt-with-direct-tool-usage](https://linear.app/rollbar-inc/issue/SDK-506/replace-grunt-with-direct-tool-usage)


waltjones (approved):
🚀

PR #1270, opened 2025-07-31
Added React 19 TS example

matux:
## Description of the change

This PR adds a React 19 TS example to the project which fails when using the `3.0.0-alpha.4` package due to CJS/ESM compatibility issues.

## Related issues

- [SDK-492/migrate-source-code-to-es-modules](https://linear.app/rollbar-inc/issue/SDK-492/migrate-source-code-to-es-modules)


examples/react-19-typescript/src/App.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import logo from './logo.svg';
+import './App.css';
+
+import Rollbar from 'rollbar';
+const instance: Rollbar = new Rollbar();
```
waltjones: Maybe a TODO that says we'll finish this up later?

matux: I mean, right now it points to the local rollbar via `file:../...`, so it works. But I can add a todo.

examples/react-19-typescript/package.json
```diff
@@ -0,0 +1,147 @@
+{
+  "name": "react-ts-18",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@babel/core": "^7.16.0",
+    "@pmmmwh/react-refresh-webpack-plugin": "^0.5.3",
+    "@rollbar/react": "../../../rollbar-react",
```
waltjones: Either remove this or use the npmjs version?

examples/react-19-typescript/src/App.tsx
matux: Oh you mean because the example usage for rollbar isn't really there. Ok, done

waltjones (approved):

PR #1288, opened 2025-08-07
Add commonjs wrappers to components when using require

matux:
## Description of the change

This PR introduces CJS wrappers for the different Rollbar components when using the browser with require.

[SDK-531/add-cjs-for-browser-components-using-require](https://linear.app/rollbar-inc/issue/SDK-531/add-cjs-for-browser-components-using-require)

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,42 +34,42 @@
     },
     "./src/browser/core": {
       "types": "./src/browser/core.d.ts",
-      "require": "./src/browser/core.js",
+      "require": "./src/browser/core.cjs",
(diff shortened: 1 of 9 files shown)


examples/webpack-require/README.md
```diff
@@ -0,0 +1,10 @@
+# Bare bones webpack ptoject
```
waltjones: Typo

matux: Done

waltjones (approved):

PR #1305, opened 2025-09-09
Export telemetry from ReplayMap

waltjones:
## Description of the change

Refactor to trigger telemetry export from ReplayMap.

## Type of change

- [x] Refactor

### Development

- [x] Lint rules pass locally

### DIFF
--- a/src/telemetry.js
+++ b/src/telemetry.js
@@ -50,6 +50,14 @@ Telemeter.prototype.copyEvents = function () {
   return events;
 };
 
+
+Telemeter.prototype.exportTelemetrySpan = function (attributes = {}) {
+  if (this.telemetrySpan) {
+    this.telemetrySpan.end(attributes);
+    this.telemetrySpan = this.tracing.startSpan('rollbar-telemetry', {});
+  }
+};
+
 Telemeter.prototype.capture = function (
   type,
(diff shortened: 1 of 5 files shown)


src/telemetry.js
```diff
@@ -50,6 +50,14 @@ Telemeter.prototype.copyEvents = function () {
   return events;
 };
 
+
+Telemeter.prototype.exportTelemetrySpan = function (attributes = {}) {
+  if (this.telemetrySpan) {
+    this.telemetrySpan.end(attributes);
+    this.telemetrySpan = this.tracing.startSpan('rollbar-telemetry', {});
+  }
+};
```
matux, line 59: I'm not sure I understand the intention, what is this function exporting and where to? Or is this just reseting/restarting the telemetry span? Or does ending the `telemetrySpan` moves data somewhere else?

waltjones: `span.end()` always causes an export. That's part of the standard tracing behavior.

This method ensures the currently open span is ended and exported so that it can be included in the next payload, and opens a new span to continue capturing telemetry events.

matux: And no new span is gonna be opened if no current telemetry span exist.

PR #1307, opened 2025-09-10
Convert Instrumenter to a class

waltjones:
## Description of the change

* Converts Instrumeter prototype functions to clas syntax.
* Removes some IE11 code.
* Converts some for loops to for ... in.
* Converts var to const/let.

## Type of change

- [x] Maintenance

### Development

- [x] Lint rules pass locally


matux:
Interesting, it seems the Angular example broke. I'm not sure why.

PR #1313, opened 2025-09-18
Enable attributes on the session, and send as span attributes on replay

waltjones:
## Description of the change

Enables attributes on the session instance, and updates the recorder to add session attributes to the replay span as span attributes.

## Type of change

### Development

- [x] Lint rules pass locally

### DIFF
--- a/src/tracing/session.js
+++ b/src/tracing/session.js
@@ -3,11 +3,14 @@ import id from './id.js';
 const SESSION_KEY = 'RollbarSession';
 
 export class Session {
+  #attributes;
+
   constructor(tracing, options) {
     this.options = options;
     this.tracing = tracing;
     this.window = tracing.window;
     this.session = null;
+    this.#attributes = {};
   }
 
   init() {
@@ -51,5 +54,13 @@ export class Session {
     }
     return this;
   }
-}
 
+  get attributes() {
+    return this.#attributes;
+  }
+
+  setAttributes(attributes) {
+    this.#attributes = { ...this.#attributes, ...attributes };
+    return this;
+  }
+}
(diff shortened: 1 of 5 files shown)


src/tracing/session.js
```diff
@@ -51,5 +54,13 @@ export class Session {
     }
     return this;
   }
-}
 
+  get attributes() {
+    return this.#attributes;
+  }
+
+  setAttributes(attributes) {
```
matux, line 62: Nit: I think you can do `set attributes` here too, instead of making a function.

##### non-blocking

test/browser.replay.recorder.test.js
```diff
@@ -148,8 +148,8 @@ describe('Recorder', function () {
 
       // Event count includes the custom end event
       expect(mockSpan.addEvent.calledThrice).to.be.true;
-      expect(mockSpan.setAttribute.calledOnce).to.be.true;
-      expect(mockSpan.setAttribute.calledWith('rollbar.replay.id', testReplayId)).to.be.true;
+      expect(mockSpan.setAttributes.calledOnce).to.be.true;
+      expect(mockSpan.setAttributes.calledWith({'rollbar.replay.id': testReplayId})).to.be.true;
```
matux, line 152: Oh, but how do you test it if you use the `set` keyword instead of making a `setAttributes` function?

matux (approved):
Beautiful

src/tracing/session.js
waltjones: The semantics are different. This method doesn't replace `attributes`, so I don't want code that looks like `attributes =`.

I'm considering `updateAttributes`.

PR #1328, opened 2025-09-24
Replay related fixes and refactoring + new tests

matux:


src/browser/replay/recorder.js
```diff
@@ -68,39 +68,36 @@ export default class Recorder {
   checkoutEveryNms() {
     // Recording may be up to two checkout intervals, therefore the checkout
     // interval is set to half of the maxSeconds.
-    return (this.options.maxSeconds || 10) * 1000 / 2;
+    return ((this.options.maxSeconds || 10) * 1000) / 2;
```
matux, line 71: For some reason either Prettier or ESLint decided to add parenthesis now. This shouldn't make any difference, but 🤷‍♂️ .

src/browser/replay/recorder.js
```diff
@@ -68,39 +68,36 @@ export default class Recorder {
...
-   * events, and then returns a payload ready for transport to the server.
+   * events, and exports the span to the tracing exporter. This is a side-effect
+   * function that doesn't return anything - the span is exported internally.
    *
    * @param {Object} tracing - The tracing system instance to create spans
-   * @param {string} replayId - Unique identifier to associate with this replay recording
-   * @returns {Object|null} A formatted payload containing spans data in OTLP format, or null if no events exist
+   * @param {Object} attributes - Attributes to add to the span
+   *  (e.g., rollbar.replay.id, rollbar.occurrence.uuid)
    */
```
matux, line 85: This function now has the same shape as Telemeter's `exportTelemetrySpan`, and instead of returning the payload (which forces us to produce all export side-effects before calling this function), it now returns nothing, making it a pure side-effect like Telemeter's `exportTelemetrySpan`.

matux: Additionally, it now accepts attributes instead of explicitly requiring to pass the occurrence uuid and the replay id, which is controlled by the `ReplayManager` now.

src/browser/replay/recorder.js
```diff
@@ -68,39 +68,36 @@ export default class Recorder {
...
-  dump(tracing, replayId, occurrenceUuid) {
-    const events = this._events.previous.concat(this._events.current);
+  exportRecordingSpan(tracing, attributes = {}) {
+    const events = this._collectEvents();
 
-    if (events.length < 2) {
-      logger.error('Replay recording cannot have less than 2 events');
-      return null;
+    if (events.length < 3) {
```
matux, line 89: `+ 1` since now we have the final ending event, which is being added in `_collectEvents` now.

src/browser/replay/recorder.js
```diff
@@ -68,39 +68,36 @@ export default class Recorder {
...
    *
    * @param {Object} tracing - The tracing system instance to create spans
-   * @param {string} replayId - Unique identifier to associate with this replay recording
-   * @returns {Object|null} A formatted payload containing spans data in OTLP format, or null if no events exist
+   * @param {Object} attributes - Attributes to add to the span
+   *  (e.g., rollbar.replay.id, rollbar.occurrence.uuid)
    */
-  dump(tracing, replayId, occurrenceUuid) {
-    const events = this._events.previous.concat(this._events.current);
+  exportRecordingSpan(tracing, attributes = {}) {
```
matux, line 86: To follow `exportTelemetrySpan`, this function now accepts `attributes` instead of explicitly requiring to pass the occurrence uuid and the replay id, which is controlled by the `ReplayManager` now as it should.

src/browser/replay/replayManager.js
```diff
@@ -42,52 +42,48 @@ export default class ReplayManager {
...
-   * and stores the result in the map using replayId as the key.
+   * Exports both telemetry and recording spans, then generates the complete payload
+   * using the tracing exporter and stores it in the map using replayId as the key.
    *
    * @param {string} replayId - The unique ID for this replay
-   * @returns {Promise<string>} A promise resolving to the processed replayId
+   * @param {string} occurrenceUuid - The UUID of the associated error occurrence
+   * @returns {string|null} The replayId if successful, or null if an error occurred
    * @private
    */
```
matux, line 54: Fixed a couple of things in this function:
1. `exportTelemetrySpan` should happen *after* exporting the recording span since the recording span export can throw if the recording happens to be invalid. If this happens, we don't want the telemetry span to end. We want to bail out from the whole situation.
2. `Recorder.dump` doesn't exist anymore and now uses the same strategy as `Telemeter.exportTelemetrySpan`, this is useful because it not only makes our code simpler and more consistent, but it allows us to control the order of exports in order to fix the problem described in 1.
3. `ReplayManager` is now in charge of calling tracing's `toPayload` to generate the payload necessary instead of `Recording`.
4. If the `_processReplay` process fails, we don't return a replay id anymore, we return `null`. This is because the replay *doesn't exist* if this function fails. So we end up with a replay id that points at nothing.
5. We don't map the `replayId` to `null` in our Map collection if processing the replay fails, since the replay doesn't exist and we have no logic to make any use of `null` replays.
6. This function isn't `async` anymore because it doesn't have to be.

src/browser/replay/replayManager.js
```diff
@@ -42,52 +42,48 @@ export default class ReplayManager {
...
-   * This method immediately returns the replayId and asynchronously processes
-   * the replay data in the background. The processing involves converting
-   * recorder events into a payload format and storing it in the map.
+   * The processing involves converting recorder events into a payload format
+   * and storing it in the map.
    *
-   * @returns {string} A unique identifier for this replay
+   * @returns {string|null} A unique identifier for this replay, or null if an
+   *  error occurred
    */
```
matux, line 80: This function now returns the replay id if we were able to add it, or `null` if we didn't (signaling the replay doesn't exist).

The catch in `_processReplay` did nothing so it was removed together with making the function non-async.

src/browser/replay/replayManager.js
```diff
@@ -103,15 +99,11 @@ export default class ReplayManager {
...
   async send(replayId) {
```
matux, line 101: The send function doesn't return either true or false, but throws if there was a problem sending the replay. This makes it much easier to control the problematic flows to discard replays that fail to send.

src/queue.js
```diff
@@ -102,12 +102,10 @@ class Queue {
...
-        (a) => a.key === 'replay_id',
-      )?.value;
+      const rid = data.attributes?.find((a) => a.key === 'replay_id')?.value;
 
-      if (replayId) {
-        item.replayId = this.replayManager.add(replayId, data.uuid);
+      if (rid) {
+        item.replayId = this.replayManager.add(rid, data.uuid);
       }
     }
```
matux, line 110: Renamed `replayId` to `rid` just because I felt the `find` was easier to read in a single line.

src/queue.js
```diff
@@ -116,15 +114,20 @@ class Queue {
...
-    } catch (e) {
+    } catch (err) {
       this._dequeuePendingRequest(data);
-      callback(e);
+
+      if (item.replayId) {
+        this.replayManager?.discard(item.replayId);
+      }
+
+      callback(err);
```
matux, line 130: I couldn't find a more graceful way to discard the replay if `makeApiRequest` throws (I wanted to control the discard from a single place: the new `_sendOrDiscardReplay` function. But, oh well.

src/queue.js
```diff
@@ -305,51 +308,54 @@ class Queue {
...
+   * is exceeded.
    *
-   * @param {string} replayId - The ID of the replay to handle
+   * @param {string} replayId - The ID of the replay to send or discard
+   * @param {Object} err - The error object from the API request, if any
    * @param {Object} response - The API response
    * @param {Object} headers - The response headers
-   * @returns {Promise<boolean>} A promise that resolves to true if replay was sent successfully,
-   *                             false if replay was discarded or an error occurred
    */
```
matux, line 345: This function is now much simpler and renamed it to make it more explicit.

Basically:
1. If we can't send the replay, discard the replay.
2. If we send the replay and fails, discard the replay.

This function is async but the promise doesn't resolve to any value since we don't need to.

src/browser/replay/replayManager.js
waltjones: `_processReplay` is async so that queue.add_item doesn't have to wait for this work to happen before sending the occurrence payload.

The main benefit of making it sync is so the id can be omitted from the occurrence payload, but the great majority of replays not sent will be due to API rate limits and account quotas, which won't be known until `send` is called. So there's limited benefit to making this sync.

matux: Right, I was too focused on error handling and I didn't see that. I understand, `_processReplay` happens asynchronously, we _assume_ the replay was exported successfully, but at the point of sending, if anything went wrong during process, we will just not find the replay now (before, we'd actually end up in an invalid state with a null replay).

I'll make `_processReplay` async again, no trouble!

matux: And the benefit you mention would be the fact that, if the processing fails, the `replayId` in `item` wouldn't be set. So the occurrence wouldn't be mapped to a replay that doesn't actually exist.

matux: It could be interesting that in those cases (which I've seen happen a lot, though not enough to justify not making this async), to actually send a replay span that's basically an error span, signaling that the replay failed to be processed. 🤔 

We talked about this, that's why I had the "error span" comment when setting the replayId to null in the map if the replay processing fails.

waltjones: Could be an error event on another span. I don't know that we'd want replay spans that aren't actually replays.

waltjones (approved):
✅ Nice improvement!

PR #1346, opened 2025-10-06
Update replay triggers and config defaults

waltjones:
## Description of the change

Replay triggers are predicates that decide whether conditions to capture and send a replay are met. A replay context is the relevant state at the time the predicates are applied. This context is matched against the configured triggers to determine whether a replay should be captured.

Trigger defaults in the config have been updated to both improve readability of the config and simplify applying the defaults in code.
```
  triggerDefaults: {
    samplingRatio: 1.0,
    preDuration: 300,
    postDuration: 5,
  },
```

This PR updates replay triggers to 
* centralize management in ReplayManager
* update structure of defaults
* add triggers for navigation and direct triggers

**Note**
The recorder is not updated to use new trigger options yet. This will be added in a follow up PR.

## Type of change

### Development

- [x] Lint rules pass locally

### DIFF
--- a/test/replay/integration/replayManager.bufferIndex.checkoutResilience.test.js
+++ b/test/replay/integration/replayManager.bufferIndex.checkoutResilience.test.js
@@ -57,7 +57,7 @@ describe('ReplayManager – Buffer Index Checkout Resilience', function () {
 
     const triggerContext = { type: 'occurrence' };
     const trigger = options.recorder.triggers[0];
-    const replayId = replayManager.capture('test-replay-id', 'test-uuid', trigger, triggerContext);
+    const replayId = replayManager.capture('test-replay-id', 'test-uuid', triggerContext);
     await clock.tickAsync(100);
 
     const cursor = replayManager._pendingLeading.get(replayId).bufferCursor;


src/browser/replay/replayPredicates.js
```diff
@@ -59,8 +70,37 @@ export default class ReplayPredicates {
...
+   * isTagMatching - Checks if the trigger's tags match any of the context item's tags.
+   * If no tags are specified in the trigger, it defaults to matching all tags.
+   * @param {Object} trigger - The trigger object containing the tags.
+   * @return {boolean} - True if the trigger's tags match any of the context item's tags, false otherwise.
+   */
+  isTagMatching(trigger, context) {
+    if (!trigger.tags || context.tags?.some(t => trigger.tags.includes(t))) {
       return true;
     }
     return false;
```
matux, line 109:
```js
!trigger.tags || context.tags?.some(t => trigger.tags.includes(t))
```
Will return `true` if `triggers.tags` exist but `context.tags` doesn't (`null`/`undefined`). Which means a trigger with specified tags will match context with no tags.

Possibly a case that may not happen, but given you're specifically using the optional chaining short-circuit, we have to assume `context.tags` could be `null`.

test/replay/integration/replayManager.bufferIndex.checkoutResilience.test.js
```diff
@@ -35,24 +47,17 @@ describe('ReplayManager – Buffer Index Checkout Resilience', function () {
...
 
     sinon.spy(recorder, 'exportRecordingSpan');
 
     recorder.start();
     await clock.tickAsync(200);
 
-    const replayId = replayManager.capture('test-replay-id', 'test-uuid');
+    const triggerContext = { type: 'occurrence' };
+    const trigger = options.recorder.triggers[0];
+    const replayId = replayManager.capture('test-replay-id', 'test-uuid', trigger, triggerContext);
```
matux: One parameter was added to `ReplayManager.capture`: `triggerContext`.

Here we're passing 2 new arguments, `trigger` and `triggerContext`, maybe we should be only passing `triggerContext`?

PR #1357, opened 2025-10-08
Add framework session attribute

waltjones:
## Description of the change

Add `rollbar.notifier.framework` session attribute.

## Type of change

### Development

- [x] Lint rules pass locally

### DIFF
--- a/src/tracing/session.js
+++ b/src/tracing/session.js
@@ -85,6 +85,7 @@ export class Session {
       'browser.mobile': navigator.userAgentData?.mobile,
       'browser.platform': navigator.userAgentData?.platform,
       'client.address': '$remote_ip', // updated at the API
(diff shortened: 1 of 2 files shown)


src/tracing/session.js
```diff
@@ -85,6 +85,7 @@ export class Session {
       'browser.mobile': navigator.userAgentData?.mobile,
       'browser.platform': navigator.userAgentData?.platform,
       'client.address': '$remote_ip', // updated at the API
+      'rollbar.notifier.framework': 'browser-js',
```
shankj3, line 88: just curious - Any reason we're doing it here and not where we set everything else notifier related?

waltjones: Doesn't rely on the config options, which is the real reason those other options are set in core.js.

I suppose when it's not a static value, that will inform where best to factor it.

PR #1362, opened 2025-10-10
Use trigger config for pre/post duration and recorder config

waltjones:
## Description of the change

Replaces recorder options `maxSeconds` and `postDuration` with trigger configs driver by the replayManager config. This allows each trigger to independently define its `preDuration` and `postDuration`. This also derives `maxPreDuration` from the trigger config, replacing `maxSeconds` in the recorder initialization.

The heart of this update is in `ReplayManager.capture`, where `trigger.preDuration` and `trigger.postDuration` for the matching trigger are now used for the current capture request.

Notes:
* Reconfiguration (`configure` method) is now driven through `ReplayManager`.
* The recorder buffers are _not_ reconfigured on `configure`, keeping the previous behavior.
* The `postDuration` behavior has been updated, but `preDuration` behavior is left as is. (Trailing duration is by default the length of the current recording buffer, and doesn't (yet) explicitly use `preDuration`.)

## Type of change

- [x] Refactor

### Development

- [x] Lint rules pass locally

### DIFF
--- a/src/telemetry.js
+++ b/src/telemetry.js
@@ -136,7 +136,7 @@ class Telemeter {
 
     // If the uuid is present, this is a message occurrence.
     if (rollbarUUID) {
-      (otelAttributes = {
+      ((otelAttributes = {
         message,
         level,
         type: 'message',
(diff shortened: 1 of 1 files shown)


src/browser/core.js
```diff
@@ -126,7 +126,7 @@ class Rollbar {
     );
 
     this.tracing?.configure(this.options);
-    this.replayManager?.recorder?.configure(this.options);
+    this.replayManager?.configure(this.options);
```
matux, line 129: 👍

PR #1371, opened 2025-10-13
Session Replay: introduce `ScheduledStreamCapture` (streaming leading chunks + precise end-of-window export) to survive multiple rrweb checkouts

matux:
> [!NOTE]
> PR Title/Description mostly generated by AI, guided by me and diff

## Description of the change

This PR replaces the single-shot leading capture with a streaming capture strategy that exports and sends leading replay data in small, sequential chunks during the postDuration window—eliminating event loss across multiple rrweb checkouts.

### Why
- With long postDuration, rrweb’s periodic checkout can overwrite early leading events before a single-shot export runs.
- Streaming chunking (< checkout interval) ensures no event ages out.
- We keep strict sequential sending (abort on first failure) to avoid rrweb gaps.

### What changed
- New controller: ScheduledStreamCapture
- Exports every `chunkMs`, queues payloads, and opportunistically sends if trailing is SENT and nothing is in flight.
- Adds a one-shot endTimer at postDuration to do a final boundary export (captures the [tail] from last cursor up to end-of-window), then flush & cleanup.
- Enforces at-most-one POST in flight per replay.
- Abort on first leading POST failure (prevents gaps); no abort for “no events” exports.
- Robust cleanup paths:
- Immediate finalize if finished && queue empty && not sending
- Finalize at end of sendIfReady() when finished === true
- discard(replayId) stops timers, clears queue, and completes.
- ReplayManager integration: uses ScheduledStreamCapture instead of ScheduledCapture; keeps existing gating:
- Gating only on trailing send (headers/quota).
- Telemetry per chunk (trailing + each leading).
- No new headers/attrs; server orders by span timestamps.

### Behavior summary
- Chunk cadence: currently 1:1 with checkoutEveryNms() (design supports clamping later).
- Final export at exact window end via endTimer (prevents “6s” stray export).
- Sequential sends: sendIfReady() drains queue FIFO; any POST failure ⇒ abort stream.
- Multiple triggers: independent controllers; no merge/pause.
- No page-unload special handling (unchanged).

### Files
- src/browser/replay/scheduledStreamCapture.js — new streaming controller
- src/browser/replay/replayManager.js — wires in streaming controller; unchanged trailing semantics
- src/browser/replay/recorder.js — unchanged (source of truth for checkout interval, cursor, collection)

### Tests

Updated/added integration tests to verify invariants (not implementation-specific call counts):
- replayManager.bufferIndex.test.js
- replayManager.bufferIndex.checkoutResilience.test.js
- Added: “sends exactly ceil(postDuration / checkoutInterval) leading chunks covering the window” (deterministic with fake timers + emitEveryNms).
- Adjusted brittle “exactly twice” assertions to focus on:
- first leading export starts at the captured cursor
- ≥1 leading POST when expected; all share same X-Rollbar-Replay-Id
- sequential sends (no overlap)
- cleanup after success/failure/discard
- monotonic cursor advancement across chunks

## Type of change

- [ ] Maintenance
- [ ] New release

## Related issues

[CAT-485/extended-post-trigger-replay-duration](https://linear.app/rollbar-inc/issue/CAT-485/extended-post-trigger-replay-duration)

### DIFF
--- a/test/replay/integration/replayManager.bufferIndex.checkoutResilience.test.js
+++ b/test/replay/integration/replayManager.bufferIndex.checkoutResilience.test.js
@@ -445,7 +445,7 @@ describe('ReplayManager - Buffer Index Checkout Resilience', function () {
     recorder._isReady = true;
     recorder._stopFn = () => {};
 
-    sinon.spy(logger, 'error');
+    sinon.spy(logger, 'debug');
 
     const triggerContext = { type: 'occurrence' };
     const replayId = replayManager.capture(
@@ -455,10 +455,10 @@ describe('ReplayManager - Buffer Index Checkout Resilience', function () {
     );
     await clock.tickAsync(100);
 
-    sinon.assert.calledOnce(logger.error);
-    expect(logger.error.firstCall.args[0]).to.be.a('string');
-    expect(logger.error.firstCall.args[1]).to.be.instanceOf(Error);
-    expect(logger.error.firstCall.args[1].message).to.equal(
+    sinon.assert.calledOnce(logger.debug);
+    expect(logger.debug.firstCall.args[0]).to.be.a('string');
+    expect(logger.debug.firstCall.args[1]).to.be.instanceOf(Error);
+    expect(logger.debug.firstCall.args[1].message).to.equal(
       'Replay recording has no events',
     );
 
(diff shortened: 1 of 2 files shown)


src/browser/replay/scheduledStreamCapture.js
```diff
@@ -0,0 +1,252 @@
...
+   * sequential sending.
+   *
+   * @param {string} replayId - The replay ID
+   * @param {string} occurrenceUuid - The occurrence UUID
+   * @param {number} postDuration - Duration in seconds to capture
+   */
+  schedule(replayId, occurrenceUuid, postDuration) {
+    const startTime = Date.now();
+    const endAt = startTime + postDuration * 1000;
+    const chunkMs = this._recorder.checkoutEveryNms();
```
matux, line 56: For now, I'm running this at exactly `checkoutEveryNms` and it seems to work fine, I did some testing with a multiplier `k ≈ 0.6...0.8`, but it resulted in more chunks, ergo more post calls.

PR #1374, opened 2025-10-14
Fix type definitions and component definitions

matux:
## Description of the change

- Updates type definition files
- Adds ReplayType
- Fixes Component type still using `recorder`
- Fixes examples that use `recorder` Component

## Type of change

- [ ] Maintenance
- [ ] New release

### DIFF
--- a/src/browser/replay/recorder.d.ts
+++ b/src/browser/replay/recorder.d.ts
@@ -11,13 +11,25 @@ export type BufferCursor = {
   offset: number;
 };
 
-export interface Recorder extends Rollbar.RecorderType {
+export class Recorder extends Rollbar.RecorderType {
+  constructor(options: Record<string, any>);
+
+  get isRecording(): boolean;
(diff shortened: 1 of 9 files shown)


src/browser/replay/recorder.d.ts
```diff
@@ -11,13 +11,25 @@ export type BufferCursor = {
...
+  checkoutEveryNms(): number;
   bufferCursor(): BufferCursor;
   exportRecordingSpan(
     tracing: any,
     attributes?: Record<string, any>,
     cursor?: BufferCursor,
   ): void;
+  start(): this;
+  stop(): this;
+  clear(): void;
```
waltjones, line 32: This file can be deleted since it's no longer the component.

matux: It's used by the IDE's own internal TS, I find it useful, that's why I didn't remove it. I did remove the cjs file, tho.

PR #1383, opened 2025-10-15
Enable scrubbing options for dom inputs

waltjones:
## Description of the change

Enables scrubbing of DOM inputs in telemetry.
* Enables `telemetryScrubber` and `scrubTelemetryInputs` as documented here: https://docs.rollbar.com/docs/rollbarjs-telemetry
* Enables replay scrubbing options when present, as described here: https://github.com/rrweb-io/rrweb/blob/master/guide.md#options

## Type of change

### Development

- [x] Lint rules pass locally

### DIFF
--- a/test/browser.telemetry.test.js
+++ b/test/browser.telemetry.test.js
@@ -76,8 +76,8 @@ describe('instrumentDom', function () {
     await loadHtml('test/fixtures/html/dom-events.html');
     scrubFields = ['foo', 'bar'];
     mask = '******';
-    ((options = { scrubFields, autoInstrument: { log: false } }),
-      (rollbar = new Rollbar({})));
+    options = { scrubFields, autoInstrument: { log: false } };
+    rollbar = new Rollbar({});
     tracing = new Tracing(window, null, {});
     tracing.initSession();
     telemeter = new Telemeter({}, tracing);


src/browser/telemetry.js
```diff
@@ -778,6 +792,51 @@ class Instrumenter {
...
+      }
+    }
+
+    // Apply replay options regardless of other scrubbing
+    if (
+      this.maskInputOptions[tagName.toLowerCase()] ||
+      this.maskInputOptions[inputType]
+    ) {
+      if (this.maskInputFn) {
+        value = this.maskInputFn(value, element);
```
matux: If `value` happens to be masked in the conditionals in line 802:813, you'd be passing the masked value to `this.maskInputFn` here.

Maybe store the value in some `originalValue` var and pass that here?

src/browser/telemetry.js
```diff
@@ -778,6 +792,51 @@ class Instrumenter {
...
+
+    if (
+      domUtil.isMatchingElement(element, this.scrubClasses, this.scrubSelectors)
+    ) {
+      value = mask;
+    }
+
+    if (inputType === 'password') {
+      value = mask;
+    }
```
matux: Could this override the return value of the custom `this.maskInputFn`? Is that expected behavior?

test/browser.telemetry.test.js
```diff
@@ -69,22 +69,19 @@ describe('instrumentNetwork', function () {
...
+  let tracing, telemeter, instrumenter, mask, options, rollbar, scrubFields;
   const wait = (t) => new Promise((resolve) => setTimeout(resolve, t));
 
   beforeEach(async function () {
     await loadHtml('test/fixtures/html/dom-events.html');
     scrubFields = ['foo', 'bar'];
-    rollbar = new Rollbar({});
+    mask = '******';
+    ((options = { scrubFields, autoInstrument: { log: false } }),
+      (rollbar = new Rollbar({})));
```
matux: ##### non-blocking, don't feel strongly, just asking

Is there any reason this is like this and not just
```js
  options = { scrubFields, autoInstrument: { log: false } };
  rollbar = new Rollbar({});
```
?

matux (commented):
This looks great! I think I found something in the `scrubInputValue` logic where I'm not sure what it does is what we want, so left two comments.

test/fixtures/html/dom-events.html
```diff
@@ -8,7 +8,7 @@
     <div class="container" id"main-container">
```
matux: Yikes, there's an `=` missing between `id` and `"main-container"` here!

src/browser/telemetry.js
waltjones: Yes, this is expected. If it's a password field, it is always masked. Making it an early return at the top would make that more obvious.

test/browser.telemetry.test.js
waltjones: I think prettier did this? Let me see what happens when I undo it.

PR #1389, opened 2025-10-16
Snippet chooses replay bundle when appropriate

matux:
## Description of the change

Updates snippet to choose replay bundle when appropriate


waltjones (approved):
This ended up in a pretty strange diff state after the merges.

PR #1473, opened 2026-06-11
PLA-1585: fix flaky undici telemetry tests + harden CI matrix

dbutts29:
## Description of the change

Fixes the flaky `with fetch network capture enabled` telemetry tests and hardens the CI matrix so a single floating-Node failure can no longer mask the whole suite. **Test-infra only — no SDK/`src` changes, zero runtime impact** (`undici` is a devDependency used only in `test/server.telemetry.test.js`; the SDK instruments the global `fetch` and never imports undici).

### Background

The two fetch-capture tests mock the network with undici `MockAgent` + `setGlobalDispatcher` but assert against Node's **global built-in `fetch`**. `MockAgent` only intercepts the built-in `fetch` when its global-dispatcher symbol matches the running Node's *bundled* undici. undici **6** uses `Symbol.for('undici.globalDispatcher.1')`; undici **7** bumped it to `.2`. So with the pinned npm `undici@6`, the mock was bypassed on any Node leg that bundles undici 7: the request escaped to the real `example.com/api/users` and returned **404** instead of the mocked **200** (a 404 rather than a `MockNotMatchedError` is the tell that it escaped the mock).

This failed **only on the `latest` leg**, but with `fail-fast` on, that one failure cancelled the other four legs — so every PR looked broken. Master had not run green since 2026-03-11, so the regression went unnoticed.

### Telemetry test fix

Measured interception across the matrix:

| Node leg | Bundled undici | npm undici 6 `MockAgent` intercepts built-in `fetch`? |
|----------|----------------|--------------------------------------------------------|
| 18 / 20 / 22 / 24 | 6.x (symbol `.1`) | ✅ yes |
| latest | 7.x (symbol `.2`) | ❌ no |

The fix keeps the tests asserting against each Node version's **real built-in `fetch`** (so the matrix still exercises per-version fetch behaviour) while making `MockAgent` intercept on every leg:

- The undecorated `undici` devDep is the **forward** version (`^7.27.2`); the **legacy** version is the `undici-v6` alias (`npm:undici@^6`).
- The test **dynamic-imports** one by Node major: `<20` → `undici-v6` (6), `>=20` → `undici` (7). undici 7's `setGlobalDispatcher` writes **both** the `.1` and `.2` slots, so the mock is seen whether the running Node bundles undici 6 (`.1`) or 7 (`.2`).

Two constraints shaped this:

1. **undici 7 requires Node 20+** — it references the global `File` constructor (added in Node 20), so it throws `ReferenceError: File is not defined` at load time on Node 18, a supported target (`CLAUDE.md`: "Minimum Node.js: 18+"). Node 18 must stay on undici 6.
2. The import therefore must be **dynamic**, so the unselected major never loads — a static forward-`undici` (7) import would crash at load on Node 18 even if unused.

**Alternatives considered and rejected:**

- *Swap `globalThis.fetch = undiciFetch`* (route global fetch through the npm undici): works on every leg, but then all legs test the npm undici's `fetch` instead of each version's real built-in `fetch` — which defeats the matrix for fetch-compatibility tests.
- *Bump everything to undici 7*: fixes `latest` but breaks the Node 18 leg (undici 7 can't load there).

### CI matrix hardening

- `fail-fast: false` — one leg's failure no longer cancels the others (the cascade that made a `latest`-only failure look universal).
- `continue-on-error` on the floating `latest` leg — keep it as a forward-compat canary that doesn't block PRs.
- `timeout-minutes: 20` plus a retry/timeout wrapper around `playwright install chromium`, which stalls indefinitely on Node 24/latest; gate the Playwright + Web Test Runner steps to stable Node (the harness Node version is irrelevant to browser coverage).

### Verification

- `npm run test:server`: **149 passing** locally on Node **18, 20, 22, and 24**; `npm run lint` clean.
- CI on `cf9742d0`: **all five legs green**, including `latest` — the undici-7-bundled environment that can't be reproduced locally, giving live confirmation the `.2` interception path works.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue) — the flaky telemetry tests
- [x] Maintenance — CI matrix hardening
- [ ] New release

## Related issues

- Fix [PLA-1585](https://linear.app/rollbar-inc/issue/PLA-1585)

### Development

- [x] Lint rules pass locally

### Code review

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/server.telemetry.test.js
+++ b/test/server.telemetry.test.js
@@ -6,12 +6,7 @@ import { URL } from 'url';
 import { expect } from 'chai';
 import nock from 'nock';
 import sinon from 'sinon';
-import {
-  fetch as undiciFetch,
-  MockAgent,
-  getGlobalDispatcher,
-  setGlobalDispatcher,
-} from 'undici';
+import { MockAgent, getGlobalDispatcher, setGlobalDispatcher } from 'undici';
 
 import Rollbar from '../src/server/rollbar.js';
 import { mergeOptions } from '../src/server/telemetry/urlHelpers.js';
@@ -272,7 +267,6 @@ describe('telemetry', function () {
     let response;
     let mockAgent;
     let originalDispatcher;
-    let originalFetch;
     let sessionId;
     let asyncLocalStorage;
 
@@ -280,14 +274,6 @@ describe('telemetry', function () {
       if (typeof fetch !== 'function') {
         this.skip();
       }
-
-      // Route the global fetch through the same undici instance that MockAgent
-      // patches. Node's built-in fetch uses the bundled undici, whose global
-      // dispatcher symbol can diverge from the npm undici across Node versions,
-      // silently bypassing setGlobalDispatcher and letting the request reach the
-      // real network.
-      originalFetch = globalThis.fetch;
-      globalThis.fetch = undiciFetch;
 
       originalDispatcher = getGlobalDispatcher();
       mockAgent = new MockAgent();
@@ -338,9 +324,6 @@ describe('telemetry', function () {
       if (originalDispatcher) {
         setGlobalDispatcher(originalDispatcher);
       }
-      if (originalFetch) {
-        globalThis.fetch = originalFetch;
-      }
     });
 
     it('message payload should have fetch telemetry', function () {
@@ -373,20 +356,11 @@ describe('telemetry', function () {
     let response;
     let mockAgent;
     let originalDispatcher;
-    let originalFetch;
 
     beforeEach(async function () {
       if (typeof fetch !== 'function') {
         this.skip();
       }
-
-      // Route the global fetch through the same undici instance that MockAgent
-      // patches. Node's built-in fetch uses the bundled undici, whose global
-      // dispatcher symbol can diverge from the npm undici across Node versions,
-      // silently bypassing setGlobalDispatcher and letting the request reach the
-      // real network.
-      originalFetch = globalThis.fetch;
-      globalThis.fetch = undiciFetch;
 
       originalDispatcher = getGlobalDispatcher();
       mockAgent = new MockAgent();
@@ -432,9 +406,6 @@ describe('telemetry', function () {
       }
       if (originalDispatcher) {
         setGlobalDispatcher(originalDispatcher);
-      }
-      if (originalFetch) {
-        globalThis.fetch = originalFetch;
       }
     });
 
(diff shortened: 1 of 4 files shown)


dbutts29:
Folded in CI-reliability changes (same PLA-1585 goal of making rollbar.js CI trustworthy):

- `fail-fast: false` — one leg's failure no longer cancels the others (the cascade that disguised this as an all-legs breakage).
- `continue-on-error` on the floating `latest` leg — kept as a forward-compat canary (it caught the undici 7.x divergence) but it can no longer block a PR or, with fail-fast off, cancel the supported legs.
- `timeout-minutes: 20` + retry/timeout around `playwright install chromium` — the chromium download stalled 40+ min on the 24/`latest` legs (a supported version, so not solved by touching `latest` alone); each attempt is now capped at 5 min with up to 3 retries.

The telemetry fix is the substantive change; these make the result observable and stop a stalled browser download from hanging the matrix.

waltjones:
Looking at using package aliasing to install two versions of undici, so the tests continue to use the version of fetch for each Node version in the test matrix.

https://docs.npmjs.com/cli/v8/using-npm/package-spec#aliases

test/server.telemetry.test.js
```diff
@@ -6,7 +6,15 @@ import { URL } from 'url';
...
-import { MockAgent, getGlobalDispatcher, setGlobalDispatcher } from 'undici';
+// MockAgent only intercepts the built-in `fetch` when its global-dispatcher
+// symbol matches the running Node's bundled undici: undici 6 (symbol `.1`) for
+// Node 18-24, undici 7 (symbol `.2`) for the undici-7-bundled `latest` leg.
+// undici 7 also requires Node 20+ (it references the global `File`), so Node 18
+// must stay on undici 6. Dynamic import so only the selected major is loaded
+// (statically importing undici 7 on Node 18 would crash at load time).
+const nodeMajor = parseInt(process.versions.node.split('.')[0], 10);
+const { MockAgent, getGlobalDispatcher, setGlobalDispatcher } =
+  nodeMajor >= 20 ? await import('undici-v7') : await import('undici');
```
waltjones: One small nit. Let's use the decorated name for the legacy version, and the undecorated name for the forward version. e.g. 'undici' and 'undici-v6'.

waltjones (approved):
One comment and then looks good.

